PR #207, opened 2025-12-26
fix(health): correct CLIProxy port detection on macOS/Linux

shunkakinoki:
## Summary

Fixes the health check incorrectly reporting "Proxy not running" and "Port 8317 free" even when CLIProxyAPI is actively running.

### Root Cause

Two bugs in `src/utils/port-utils.ts`:

1. **Mutually exclusive lsof flags**: The command `lsof -i :${port} -sTCP:LISTEN -t -F pcn` fails because `-t` and `-F` are mutually exclusive in lsof. This causes the command to error, and the catch block assumes "port is free".

2. **Missing process names**: The `isCLIProxyProcess` whitelist was missing `cliproxyapi` and `cli-proxy-api-plus` (common binary names).

### Changes

- Removed `-t` flag from lsof command (line 41)
- Added `cliproxyapi`, `cliproxyapi.exe`, `cli-proxy-api-plus`, `cli-proxy-api-plus.exe` to process name whitelist

## Test plan

- [ ] Run `ccs doctor` or check dashboard health when CLIProxyAPI is running
- [ ] Verify "CLIProxy Port" now shows "CLIProxy running" with correct PID

### DIFF
--- a/src/utils/port-utils.ts
+++ b/src/utils/port-utils.ts
@@ -111,26 +111,17 @@ async function getPortProcessWindows(port: number): Promise<PortProcess | null> 
 
 /**
  * Check if process is CLIProxy
+ * Uses prefix matching to handle Linux kernel's 15-char process name truncation
+ * (e.g., 'cli-proxy-api-plus' becomes 'cli-proxy-api-p' in lsof/ps output)
  */
 export function isCLIProxyProcess(process: PortProcess | null): boolean {
   if (!process) {
     return false;
   }
 
-  // Match various CLIProxy binary names
   const name = process.processName.toLowerCase();
-  return [
-    'cli-proxy',
-    'cli-proxy.exe',
-    'cliproxy',
-    'cliproxy.exe',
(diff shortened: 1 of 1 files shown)


kaitranntt:
## Review Update

Added a follow-up commit to address Linux kernel process name truncation:

**Issue**: Linux truncates process names to 15 characters. `cli-proxy-api-plus` (18 chars) becomes `cli-proxy-api-p` in lsof/ps output, which wasn't in the whitelist.

**Fix**: Switched from exact whitelist matching to prefix-based matching:
```typescript
// Before: ['cli-proxy', 'cli-proxy-api-plus', ...].includes(name)
// After:  name.startsWith('cli-proxy') || name.startsWith('cliproxy')
```

**Benefits**:
- Handles OS-level truncation automatically
- Future-proof for new binary names
- Simpler code (-9 lines)

The original lsof `-t` flag fix is correct. This commit builds on it to ensure detection works for all CLIProxy variants.

PR #330, opened 2026-01-14
fix(cliproxy): return null for unknown quota, add verbose diagnostics

kaitranntt:
## Summary
- Fix `calculateAverageQuota()` returning 100% when quota fetch fails or returns no model data
- Display "N/A (fetch unavailable)" instead of misleading 100% for unknown quota
- Add `--verbose` flag to quota-related functions for troubleshooting quota fetch issues

## Changes
- **quota-manager.ts**: Return `null` instead of `100` when no model data, update callers with null coalescing
- **cliproxy-executor.ts**: Show "N/A (fetch unavailable)" for null quota values
- **quota-fetcher.ts**: Add verbose parameter with debug logging for token refresh and API responses
- **cliproxy-command.ts**: Pass verbose flag through quota functions

## Test plan
- [x] `bun run validate` passes (typecheck, lint, format, tests)
- [ ] `ccs cliproxy --verbose` shows quota fetch diagnostics
- [ ] Quota display shows "N/A" when fetch fails instead of 100%

### DIFF
--- a/src/cliproxy/cliproxy-executor.ts
+++ b/src/cliproxy/cliproxy-executor.ts
@@ -485,8 +485,10 @@ export async function execClaudeWithCLIProxy(
     if (preflight.switchedFrom) {
       console.log(info(`Auto-switched to ${preflight.accountId}`));
       console.log(`    Reason: ${preflight.reason}`);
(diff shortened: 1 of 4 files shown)


kaitranntt:
Recreating PR

PR #676, opened 2026-03-03
fix(cliproxy): reduce gemini alias model bloat

kaitranntt:
## Summary
- stop aggressive Gemini alias expansion in CLIProxy config generation
- remove cache-derived Antigravity alias enrichment that inflated model listing
- keep lightweight alias normalization only (dot/hyphen + customtools toggle)
- bump cliproxy config schema marker to force regeneration
- update unit test expectations to lock the non-enrichment behavior

## Validation
- bun run build
- bun test tests/unit/cliproxy/config-generator.test.js
- pre-push CI parity gate (build + validate + tests)


kaitranntt:
Addressed reviewer suggestion in follow-up commit e28d9c8a:\n- strengthened the non-enrichment regression test to also assert baseline Gemini aliases remain present after regeneration.

PR #726, opened 2026-03-15
feat: add support for setting the global fetch proxy through HTTP_PROXY

### LINKED ISSUE #714: Question: setting HTTP_PROXY to take effect in ccs-dashboard web page
On the provider section of the CCS Dashboard web homepage, the quota for each account is showing “fetched failed.” Could this be caused by the need to configure HTTP_PROXY? 
I’m able to view the quota for the same account in cliproxyplus (with HTTP_PROXY in settings), so I’m wondering whether this is a dashboard-side proxy or network configuration issue.

Kherrisan:
feat: add support for setting the global fetch proxy through HTTP_PROXY

This PR adds standard fetch proxy support for CCS dashboard and provider-management requests.

Maintainer follow-up commits harden the original change so CCS loopback traffic stays direct, invalid proxy env values do not break startup, proxy precedence matches downloader behavior, and runtime fetch calls use the configured dispatcher.

Closes #714

### DIFF
--- a/src/utils/fetch-proxy-setup.ts
+++ b/src/utils/fetch-proxy-setup.ts
@@ -5,8 +5,10 @@
  * variables without routing CCS loopback traffic back through the proxy.
  */
 
-import { Agent, Dispatcher, ProxyAgent, setGlobalDispatcher } from 'undici';
+import { Agent, Dispatcher, ProxyAgent, fetch as undiciFetch, setGlobalDispatcher } from 'undici';
 import { getProxyResolution, shouldBypassProxy } from './proxy-env';
+
+const FETCH_PROXY_PROTOCOLS = ['http:', 'https:'];
 
 type GlobalFetchProxyConfig = {
   httpProxyUrl?: string;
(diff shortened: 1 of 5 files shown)


kaitranntt:
Pushed a maintainer follow-up onto this branch to keep the original feature intact and make it safer to merge.

What changed in the follow-up commit:
- kept the global fetch proxy support from this PR
- bypassed CCS loopback/local management traffic even when `NO_PROXY` is unset
- fail-open on malformed proxy env instead of crashing startup
- added `ALL_PROXY` fallback and aligned proxy precedence with the downloader path
- shared proxy env parsing between fetch and downloader logic
- added focused tests and README notes
- moved `undici` back to the v5 line so the repo still matches its declared `node >=18.0.0` support

Validation on the maintainer patch:
- `bun run typecheck`
- focused proxy test suite
- `bun run build`
- `bun run validate`
- smoke test with invalid `HTTP_PROXY`

The only non-green gate left locally was `validate:ci-parity`, and that was because this PR head branch is behind `origin/dev`, not because of the proxy changes themselves.

kaitranntt:
Follow-up maintainer pass pushed after deeper review.

This second commit tightened two things that mattered for merge safety:
- preserved existing SOCKS-style proxy env behavior for the downloader path instead of narrowing everything to http(s)
- explicitly rebinds global fetch when proxying is enabled so CCS runtime fetch call sites use the configured dispatcher

Also added deeper coverage for restricted-protocol fallback resolution, automatic loopback bypass branches, fetch routing decisions, and NO_PROXY handling.

PR #790, opened 2026-03-25
fix(docker): add source label to link GHCR package to repository

kapdon:
## Summary
- Adds `org.opencontainers.image.source` label directly in the Dockerfile so GHCR links the package to the repository
- Without this label baked into the image, the package only appears under the user's profile Packages tab instead of the repo's Packages tab
- The `labels` parameter in `docker/build-push-action` sets OCI manifest annotations but GHCR requires the label embedded in the image itself for repo linking

## Test plan
- [ ] Verify next stable release triggers `docker-release.yml` and the resulting image appears under the repo's Packages tab
- [ ] Confirm `docker inspect` on the published image shows the `org.opencontainers.image.source` label

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -40,6 +40,8 @@ RUN test -d dist && test -d lib && echo "[OK] Build artifacts validated"
 # =============================================================================
 FROM node:20-bookworm-slim AS runtime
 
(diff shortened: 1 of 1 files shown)


kapdon:
Closing — the `docker/build-push-action` labels parameter already embeds the source label in the image. The actual fix is to manually link the package to the repo in GHCR settings. Opening an issue instead.

PR #792, opened 2026-03-25
fix(docker): add index annotations to link GHCR package to repository

kapdon:
## Summary
- Multi-platform builds create a manifest index pointing to per-platform images
- The existing `labels` parameter only sets labels on per-platform image configs, but GHCR reads annotations from the **manifest index** for repository auto-linking
- Adds `index:`-prefixed `annotations` so the package appears under the repo Packages tab instead of only the user profile

## Test plan
- [ ] Verify next stable release triggers `docker-release.yml` and the resulting package appears under the repo's Packages tab
- [ ] `docker buildx imagetools inspect ghcr.io/kaitranntt/ccs-dashboard:latest` should show `org.opencontainers.image.source` in the index annotations

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -121,5 +121,8 @@ jobs:
             org.opencontainers.image.source=https://github.com/${{ github.repository }}
             org.opencontainers.image.version=${{ steps.meta.outputs.version }}
             org.opencontainers.image.revision=${{ steps.meta.outputs.revision }}
+          annotations: |
+            index:org.opencontainers.image.source=https://github.com/${{ github.repository }}
+            index:org.opencontainers.image.description=CCS Dashboard container image
           cache-from: type=gha
           cache-to: type=gha,mode=max


kapdon:
@kaitranntt weird, I think because the package is already created you might have to manually connect it to this repo from your package section, the rebuild succeeded but it still is only showing up under your username and not part of the repo.

Profile/Org → Packages → package → Connect repository. GitHub documents that flow directly.

kaitranntt:
@kapdon Follow-up after checking the package state:

- `ccs-dashboard` now shows `kaitranntt/ccs` as its repository source
- `org.opencontainers.image.source` is set to `https://github.com/kaitranntt/ccs`
- the package resolves under `https://github.com/kaitranntt/ccs/pkgs/container/ccs-dashboard`

So PR #792's `index:` annotation fix does appear to be in effect now.

What I still can't tell from the current state is whether GitHub updated the association automatically after republish, or whether it needed a manual `Connect repository` step because the package already existed.

Do you know if this got fixed just by republishing after the annotation change, or did someone manually use `Connect repository` at some point?

kapdon:
mmm no idea, maybe since github was having such uptime issues as they are about to lose their last 9 in uptime. it was delayed after the annotation change because it wasn't showing up before, since no one can manually run the connect repository function besides you the owner, so if you didn't do it then maybe annotation just took awhile to propagate, it works now though!

package now shows up on sidebar

image

PR #870, opened 2026-04-01
fix: probe Codex config override support directly

### LINKED ISSUE #869: bug: Codex config override probe rejects supported CLI builds
## Description
CCS can reject native Codex launches with `[X] Codex CLI does not advertise --config overrides...` even when the installed Codex binary supports `-c/--config`.

## Root Cause
`src/targets/codex-detector.ts` currently infers support from a brittle `--help` substring match instead of probing the actual flag behavior.

## Affected Files
- src/targets/codex-detector.ts
- src/targets/codex-adapter.ts
- src/web-server/services/codex-dashboard-service.ts
- tests/unit/targets/codex-detector.test.ts
- tests/unit/targets/codex-adapter.test.ts
- tests/unit/targets/codex-runtime-integration.test.ts

kaitranntt:
## Summary
- probe native Codex `-c/--config` support by executing `codex -c 'model="gpt-5"' --version` before trusting help output
- keep a broader help-text fallback for builds where the direct probe fails but the flag is still advertised
- add regression coverage for detector, runtime launch, bridge launch, and dashboard warning paths

## Testing
- bun test tests/unit/targets/codex-detector.test.ts
- bun test tests/unit/targets/codex-adapter.test.ts
- bun test tests/unit/targets/codex-runtime-integration.test.ts
- bun test tests/unit/targets/codex-settings-bridge-launch.test.ts
- bun test tests/unit/web-server/codex-dashboard-service.test.ts
- bun run format
- bun run lint:fix
- bun run build
- bun run validate
- bun run validate:ci-parity

Closes #869

Docs impact: none
Action: no update needed — detector and test behavior only; no user-facing docs or help flow changed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     "ui:build": "cd ui && bun run build",
     "ui:preview": "cd ui && bun run preview",
     "ui:validate": "cd ui && bun run validate",
+    "prepack": "bun run build:all",
     "prepare": "husky",
     "postinstall": "node scripts/postinstall.js"
   },
--- a/src/targets/codex-adapter.ts
+++ b/src/targets/codex-adapter.ts
@@ -2,6 +2,7 @@ import { ChildProcess, spawn } from 'child_process';
 import * as fs from 'fs';
 import type { ProfileType } from '../types/profile';
 import { runCleanup } from '../errors';
+import { expandPath } from '../utils/helpers';
 import { wireChildProcessSignals } from '../utils/signal-forwarder';
 import { escapeShellArg, stripAnthropicEnv, stripCodexSessionEnv } from '../utils/shell-executor';
 import type {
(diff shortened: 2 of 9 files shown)


kaitranntt:
Follow-up vanilla audit on the packaged branch using the maintainer machine's existing `.ccs/cliproxy` auth, but a fresh Codex home.

Matrix executed:
- packaged install via `npm install -g <tgz>` into a temp prefix
- copied current `~/.ccs` into an isolated `CCS_HOME`
- fresh `HOME` / `CODEX_HOME` with no prior Codex config
- real native Codex binary on PATH (`codex-cli 0.119.0-alpha.1`)

Observed before this follow-up commit:
- `ccsxp exec ...` failed when `CODEX_HOME` was explicitly set to a path that did not exist yet
- the failure came from Codex before request execution (`Error finding codex home`)

Observed after this follow-up commit:
- `ccsxp exec --skip-git-repo-check --json ...` succeeds with a fresh explicit `CODEX_HOME`
- `ccs codex --target codex exec --skip-git-repo-check --json ...` succeeds with the same fresh setup
- plain fresh `codex exec ...` still fails against the default OpenAI route in that environment, which confirms the CCS path is removing real setup burden rather than just matching native behavior

Remaining note from the audit:
- if `HOME` itself lives under a macOS temp dir, native Codex warns about refusing to create helper binaries there; that is a test-harness artifact, not a normal user-home behavior. Future manual smoke scripts should use a non-temp isolated home path.

PR #963, opened 2026-04-11
test(ci): isolate CCS-managed env vars from host session

sgaluza:
## Summary

- Fix test isolation when running inside a CCS-managed session
- `CLAUDE_CONFIG_DIR` and `CCS_PROFILE_TYPE` leak from host CCS sessions into test processes, causing spurious failures
- Three test files patched: `persist-command-handler`, `claudecode-env-stripping`, `websearch-transformer`

## Changes

- `tests/unit/commands/persist-command-handler.test.ts` — delete `CLAUDE_CONFIG_DIR` in beforeEach so scoped `CCS_HOME` takes effect
- `tests/unit/utils/claudecode-env-stripping.test.ts` — full save/delete/restore cycle for `CLAUDE_CONFIG_DIR`
- `tests/unit/hooks/websearch-transformer.test.ts` — add `CCS_PROFILE_TYPE: ''` to all 6 subprocess env blocks to prevent `shouldSkipHook()` from calling `process.exit(0)`

## Test plan

- [x] `bun run validate` passes
- [x] All 3 modified test files pass when run inside a CCS session
- [x] No production code modified — test-only changes

### DIFF
--- a/tests/unit/commands/persist-command-handler.test.ts
+++ b/tests/unit/commands/persist-command-handler.test.ts
@@ -73,6 +73,7 @@ function stubProcessExit(): void {
 beforeEach(async () => {
   tempRoot = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'ccs-persist-handler-test-'));
   originalClaudeConfigDir = process.env.CLAUDE_CONFIG_DIR;
+  delete process.env.CLAUDE_CONFIG_DIR;
   originalProcessExit = process.exit;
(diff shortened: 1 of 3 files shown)


kaitranntt:
Thanks for the clean fix here. This addresses the leaked CCS session env problem directly, the isolation changes are scoped well, and the targeted tests reproduce the failure mode clearly. I’m grateful for the careful follow-up on the review feedback as well. I’m going to merge this.

PR #975, opened 2026-04-13
fix(image-analysis): wrap defaultDeps in arrow functions to fix circular dep capture

### LINKED ISSUE #973: bug: image-analysis readiness falls back to native Read after `deps.initializeAccounts` runtime error
## Summary

CCS image-analysis is not usable for me right now.
During a Codex-backed CCS session, readiness falls back to native `Read` after this runtime warning:

```text
[i] CCS could not verify OpenAI Codex auth readiness: deps.initializeAccounts is not a function This session will use native Read.
```

So the session degrades to native `Read` instead of using the CCS image-analysis path.

## Actual behavior

- CCS attempts to verify image-analysis auth readiness
- readiness check throws:
  - `deps.initializeAccounts is not a function`
- CCS falls back to native `Read`
- result: CCS-specific image-analysis is unavailable in the session

## Expected behavior

- readiness check should complete normally
- if auth is actually missing, CCS should report a normal auth-missing state
- it should not fail with an internal runtime error and silently degrade the feature

## Environment

Checked local environment:

- OS: Linux
- Node.js: `v20.19.6`
- active `ccs` binary:
  - `/home/node/.nvm/versions/node/v20.19.6/bin/ccs`
- `which ccs`:
  - `/home/node/.nvm/versions/node/v20.19.6/bin/ccs`
- `/home/node/.nvm/versions/node/v20.19.6/bin/ccs --version` reports:
  - CCS `v7.68.1`
  - active install location: `/home/node/.nvm/versions/node/v20.19.6/bin/ccs`
- `bun --bun "$HOME/.bun/bin/ccs" --version` reports:
  - CCS `v7.68.1`
  - install location: `/home/node/node_modules/@kaitranntt/ccs/dist/ccs.js`

So I re-checked after your comment and the currently active installs I tested are reporting `7.68.1`, not `7.55.0`.

## Reproduction example

```bash
/home/node/.nvm/versions/node/v20.19.6/bin/ccs codex --effort xhigh --model 'gpt-5.4(xhigh)[1m]' --allow-dangerously-skip-permissions --dangerously-skip-permissions
```

Observed startup excerpt:

```text
[OK] WebSearch: Ready (DuckDuckGo)
[OK] CLIProxy binary ready (0.0s)
[i] Joined existing CLIProxy on port 8317 (http)
[i] CCS could not verify OpenAI Codex auth readiness: deps.initializeAccounts is not a function This session will use native Read.
[i] Thinking: xhigh (flag: --effort xhigh)
```

The session then continues, but image-analysis falls back to native `Read`.

## Auth-related context

ChatGPT OAuth Codex has already been added on my side and there is already a Codex account present, so this does not look like a first-time auth/bootstrap case.

## Why I still think this is a bug

The observed failure is not a normal auth result such as:

- auth missing
- token missing
- not authenticated

Instead it is an internal runtime error:

- `deps.initializeAccounts is not a function`

So the feature appears to fail before readiness can be evaluated correctly.

## User impact

- CCS image-analysis path becomes unavailable
- session degrades to native `Read`
- user can still read images via native Claude vision, but cannot use the CCS-specific image-analysis path

innocarpe:
## Problem

Fixes #973

`deps.initializeAccounts is not a function` (and `deps.getAuthStatus is not a function`) errors occur even on v7.68.1. The issue persists after reinstalling because the root cause is a **circular dependency in the module graph**, not a stale installation.

## Root Cause

In `defaultDeps`, most function references are captured directly at module evaluation time:

```ts
const defaultDeps: ImageAnalysisRuntimeStatusDeps = {
  checkRemoteProxy,       // captured at module load time
  fetchRemoteAuthStatus,  // captured at module load time
  getAuthStatus,          // captured at module load time
  getProxyTarget,         // captured at module load time
  initializeAccounts,     // captured at module load time
  isCliproxyRunning: () => isCliproxyRunning(),  // ✅ already wrapped
};
```

When `cliproxy/executor/index.js` (or similar entry points) loads this module, `auth-handler.js` is mid-initialization due to circular dependencies. At that moment, `initializeAccounts`, `getAuthStatus`, etc. are still `undefined`, and those `undefined` values get locked into `defaultDeps`.

`isCliproxyRunning` was already correctly wrapped in an arrow function — deferring evaluation to call time, by which point all modules are fully initialized.

## Fix

Apply the same arrow function wrapping consistently to all deps:

```ts
const defaultDeps: ImageAnalysisRuntimeStatusDeps = {
  checkRemoteProxy: (...args) => checkRemoteProxy(...args),
  fetchRemoteAuthStatus: (...args) => fetchRemoteAuthStatus(...args),
  getAuthStatus: (...args) => getAuthStatus(...args),
  getProxyTarget: () => getProxyTarget(),
  initializeAccounts: () => initializeAccounts(),
  isCliproxyRunning: () => isCliproxyRunning(),
};
```

## Verification

- Reproduced the error on v7.68.1 by tracing the module load order with Node.js instrumentation
- Confirmed `auth_handler_1.initializeAccounts` is `undefined` when first captured, but becomes a function after full initialization
- Applied the patch locally — error no longer occurs
- All 6 existing unit tests pass

### DIFF
--- a/src/utils/hooks/image-analysis-runtime-status.ts
+++ b/src/utils/hooks/image-analysis-runtime-status.ts
@@ -29,11 +29,11 @@ interface ImageAnalysisRuntimeStatusDeps {
 }
 
 const defaultDeps: ImageAnalysisRuntimeStatusDeps = {
-  checkRemoteProxy,
-  fetchRemoteAuthStatus,
-  getAuthStatus,
-  getProxyTarget,
-  initializeAccounts,
+  checkRemoteProxy: (...args) => checkRemoteProxy(...args),
+  fetchRemoteAuthStatus: (...args) => fetchRemoteAuthStatus(...args),
+  getAuthStatus: (...args) => getAuthStatus(...args),
+  getProxyTarget: () => getProxyTarget(),
+  initializeAccounts: () => initializeAccounts(),
   isCliproxyRunning: () => isCliproxyRunning(),
 };
 
--- a/tests/unit/utils/hooks/image-analysis-runtime-status-circular-dependency.test.ts
+++ b/tests/unit/utils/hooks/image-analysis-runtime-status-circular-dependency.test.ts
@@ -1,0 +1,144 @@
+import { describe, expect, it } from 'bun:test';
+import { readFileSync } from 'fs';
+import { dirname } from 'path';
+import type { ImageAnalysisStatus } from '../../../../src/utils/hooks/image-analysis-backend-resolver';
+import type { AuthStatus } from '../../../../src/cliproxy/auth-handler';
+import {
+  ModuleKind,
+  ScriptTarget,
+  transpileModule,
+  type CompilerOptions,
+} from 'typescript';
+
+function createStatus(overrides: Partial<ImageAnalysisStatus> = {}): ImageAnalysisStatus {
+  return {
+    enabled: true,
+    supported: true,
+    status: 'active',
+    backendId: 'ghcp',
+    backendDisplayName: 'GitHub Copilot (OAuth)',
+    model: 'claude-haiku-4.5',
+    resolutionSource: 'profile-backend',
+    reason: null,
+    shouldPersistHook: true,
+    persistencePath: '/tmp/orq.settings.json',
+    runtimePath: '/api/provider/ghcp',
+    usesCurrentTarget: true,
+    usesCurrentAuthToken: true,
+    hookInstalled: true,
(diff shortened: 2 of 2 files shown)


kaitranntt (approved):
Thanks for the clear fix and root-cause write-up. The earlier AI review failure was due to model infrastructure, not an issue with this patch. I reviewed this manually, added regression coverage for the CommonJS circular-dependency case, and everything looks good locally. LGTM, ready to merge.

Issue #990, opened 2026-04-14
bug: ccsxp Windows regression: browser MCP args saved as string → config parse failure, downgrade exposes Codex override error


Runan92:
### Affected area

CLI runtime

## Summary
  After upgrading CCS to 7.69.1 on Windows, every `ccsxp` run immediately fails with:

  Error loading config.toml: invalid type: string "[-y,@playwright/mcp@0.0.70]", expected a sequence in mcp_servers.ccs_browser.args

  The config on disk still has `args = ["-y", "@playwright/mcp@0.0.70"]`, so CCS itself is rewriting the overrides as a TOML string when launching Codex through the PowerShell wrapper. I tried letting an assistant “auto-fix” it but the error persisted, so I downgraded to 7.68.1. Unfortunately, after the downgrade CCS still crashed with:

  [X] Codex CLI does not advertise --config overrides. Upgrade Codex before using CCS-backed Codex profiles or runtime reasoning overrides.

  The downgrade only became usable again once I reverted all changes the assistant had made to the local CCS files/environment.

  ## Environment
  - Windows 10
  - CCS 7.69.1 (upgrade) → 7.68.1 (downgrade)
  - Codex CLI 0.120.0 (`npm install -g @openai/codex`)
  - Codex path: `\AppData\Local\nodejs-lts\codex.cmd`

  ## Steps to Reproduce
  1. Install CCS 7.69.1 on Windows with Codex CLI 0.120.0 available.
  2. Open a fresh PowerShell window and run:
     ```powershell
     ccsxp

  3. Observe the config parse error.
  4. Downgrade to CCS 7.68.1 and repeat the run.
  5. Observe the does not advertise --config overrides error until all local CCS edits are reverted.

  ## Expected Result

  ccsxp launches Codex normally.

  ## Actual Result

  - 7.69.1: CCS injects browser MCP args as a string, so the Codex config parser aborts before Codex even starts.
  - 7.68.1 (after downgrade): CCS continues to reject the same Codex CLI with an incorrect --config overrides warning until local modifications are reverted.

  ## Notes

  - The behavior regressed only after the 7.69.1 upgrade.
  - Manual fixes and reverting edits were required to get the downgraded CCS working again.

### Reproduction steps

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### CCS version

_No response_

### Operating system

None

### Environment details

_No response_

### Logs, screenshots, or terminal output

```shell

```

### Before submitting

[...]

Issue #871, opened 2026-04-01
bug: Windows: ccs codex --target codex falsely reports "does not advertise --config overrides" unless CCS_CODEX_PATH is forced to codex.ps1


Runan92:
### Affected area

CLI runtime

## Summary

On Windows, `ccs codex --target codex` fails with:

```text
[X] Codex CLI does not advertise --config overrides. Upgrade Codex before using CCS-backed Codex profiles or runtime reasoning overrides.
```

But the installed Codex CLI does support `--config`, and the command works if I force CCS to use the PowerShell wrapper instead of the default detected binary.

## Workaround

This works:

```powershell
$env:CCS_CODEX_PATH = "<global-node-bin>\\codex.ps1"
ccs codex --target codex "hello"
```

So forcing `codex.ps1` avoids the issue.

## Suspected cause

This looks like a Windows-specific detection/probing issue in CCS.

It appears CCS prefers `codex.cmd` on Windows and probes via `cmd.exe`, while forcing `codex.ps1` works.

The detector also seems to swallow probe failures and then surfaces a generic message about missing `--config` support, which is misleading in this case because Codex clearly advertises `--config`.

## Notes

This was reproduced after fixing PATH resolution so that only the npm/global Codex CLI remained visible, so it does not appear to be caused by a Windows app alias collision.

## Repro

1. Install Codex CLI globally.
2. Verify Codex supports config overrides:

```powershell
codex --help
```

The output includes:

```text
-c, --config <key=value>
```

3. Run:

```powershell
ccs codex --target codex "hello"
```

## Expected

CCS should detect that Codex supports `--config` and launch normally.

## Actual

CCS fails with:

```text
[X] Codex CLI does not advertise --config overrides. Upgrade Codex before using CCS-backed Codex profiles or runtime reasoning overrides.
```

### CCS version

7.63.0

### Operating system

Windows

## Environment

- OS: Windows
- CCS: v7.63.0
- Codex CLI: `codex-cli 0.118.0`

### Logs, screenshots, or terminal output

```shell

```

### Before submitting

- [x] I searched existing issues first.
- [x] I removed secrets and private data from logs/screenshots.
- [x] I can still reproduce this on the latest released or dev build.

kaitranntt:
This is fixed in `v7.64.0`.

[...]

PR #1036, opened 2026-04-17
fix(codex): probe Windows cmd wrappers via cmd shell

Runan92:
## Summary
Fix Windows Codex detector probing for `.cmd` wrappers.

## Problem
On Windows, `ccsxp` can fail with:

`Codex CLI does not advertise --config overrides`

even when the installed Codex wrapper does support `--config`.

## Root cause
`v7.72.1` can prefer `codex.cmd` on some Windows PATH layouts, but the detector probes `.cmd/.bat` wrappers through a `cmd.exe /c` path that misquotes the wrapper command on this setup. The failed probe makes CCS think Codex lacks config override support.

## Fix
Use `spawnSync(cmdString, { shell: 'cmd.exe' })` for `.cmd/.bat` probes so the wrapper is executed with the quoting contract expected by `escapeShellArg()`.

## Validation
- Reproduced on Windows local machine
- Verified patched install resolves the `ccsxp` failure
- `node node_modules/typescript/bin/tsc --noEmit` passes
- `bun test tests/unit/targets/codex-detector.test.ts` is still blocked locally by a Bun temp-dir cleanup permission issue on Windows

### DIFF
--- a/src/targets/codex-detector.ts
+++ b/src/targets/codex-detector.ts
@@ -53,12 +53,14 @@ function runCodexProbe(codexPath: string, args: string[]): string | undefined {
 
     if (needsShell) {
       const cmdString = [codexPath, ...args].map(escapeShellArg).join(' ');
-      return childProcess.execFileSync('cmd.exe', ['/d', '/s', '/c', cmdString], {
+      const result = childProcess.spawnSync(cmdString, {
         encoding: 'utf8',
         stdio: ['ignore', 'pipe', 'ignore'],
         timeout: 5000,
         windowsHide: true,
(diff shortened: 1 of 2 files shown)


kaitranntt:
Merged into `dev`.

Thanks @Runan92 for staying on this across the original Windows reports, the follow-up validation in #871 and #990, and for bringing back a focused repro when the issue was only partially fixed. Thanks @namnh2-RS as well for the detector patch itself.

This was a useful catch: the remaining gap was specifically in how CCS probed Windows `codex.cmd` wrappers, not just in wrapper selection. Appreciate the persistence and the clean, targeted fix.

PR #1037, opened 2026-04-18
feat(proxy): enhance Anthropic-to-OpenAI message transformation

grndis:
## Summary
- support adaptive thinking mode and strip thinking blocks from history
- normalize OpenAI tool schemas and coalesce messages


grndis:
During my test, the request was rejected by the upstream.
```
❯ hello
  ⎿  Retrying in 2s · attempt 4/10 · API_TIMEOUT_MS=600000ms, try increasing it
```

PR #1047, opened 2026-04-19
fix(repo): remove tracked plans workspace files

kaitranntt:
## Summary
- remove the force-added tracked `plans/20260415-1016-cursor-provider-legacy-split` subtree from Git while leaving local ignored plan files intact
- fail the pre-push CI parity gate if anything under `plans/` is tracked again

## Root cause
`plans/` is already ignored in this repo, but commit `3b5941c60bf5578c49785c001f85612cfec8a7e5` force-added one ignored plan subtree as part of a larger feature commit.

## Verification
- `CCS_PR_BASE=main bash scripts/ci-parity-gate.sh`


kaitranntt:
Closing this because the cleanup already shipped on dev and we are promoting dev to main instead of carrying a separate main-only cleanup PR.

PR #1053, opened 2026-04-20
fix(proxy): allow openai-compatible profiles to run on separate local ports

innocarpe:
## Summary

This PR fixes a structural limitation in the OpenAI-compatible local proxy path used by CCS settings profiles.

Today, the proxy lifecycle is effectively global:

- the daemon PID/session state is stored in a single shared location
- the startup path treats the running proxy as a singleton
- the default lifecycle assumes one shared local port
- starting a second OpenAI-compatible settings profile is rejected even when a different port would be valid

That design causes real profile collisions for aliases that should remain independent. In practice, profiles such as `ccg` and `ccgm` can end up contending for the same local proxy runtime and port, even though their settings and intended routing differ.

This change makes the OpenAI-compatible proxy lifecycle profile-scoped while preserving the existing UX for the common single-profile case.

As a small repo hygiene follow-up, this PR also updates `.gitignore` to ignore local-only agent override files so personal Codex/Claude guidance does not get accidentally staged into future commits.

## Problem Background

The issue here is not just that `3456` was the default port.

The deeper problem was that CCS modeled the OpenAI-compatible proxy as a single global daemon:

- proxy state lived in one PID file and one session file
- `startOpenAICompatProxy()` would treat any running proxy as the active instance
- if that running instance belonged to a different profile, CCS refused to start a second one instead of allocating another port
- `ccs proxy status` / `activate` / `stop` also assumed one active proxy

Because of that structure, different OpenAI-compatible settings profiles were forced into one shared runtime lane. That is exactly the failure mode behind collisions like:

`Proxy already running for profile "gpt-oauth-mini" on port 3456. Stop it before starting a different profile.`

From a user perspective, this means profile separation breaks down specifically in the local proxy bridge layer.

## Root Cause

The collision came from a combination of behaviors in the existing implementation:

1. `OPENAI_COMPAT_PROXY_DEFAULT_PORT` was a single shared default.
2. PID/session files were global rather than profile-specific.
3. The daemon lifecycle only tracked one running proxy at a time.
4. Starting a different profile while one proxy was already running returned an error instead of creating another isolated proxy instance.
5. There was no first-class place in unified config to express profile-specific proxy port preferences.

So even though the CLI accepted `--port`, the lifecycle itself was still singleton-oriented.

### 1. Profile-scoped proxy state

The proxy daemon state is now stored per profile rather than globally.

Instead of a single:

- `daemon.pid`
- `session.json`

CCS now stores profile-specific daemon/session files under the proxy state directory.

This is the key structural change that allows multiple OpenAI-compatible profile proxies to coexist safely.

### 2. Profile-aware lifecycle management

The daemon lifecycle now:

- reads status per profile
- reuses the existing daemon only when the same profile is already running on the same bind target
- stops/restarts only that profile's daemon when needed
- allows another profile to start on a different port instead of failing simply because some other profile is already active

This keeps same-profile behavior intuitive while removing the cross-profile singleton constraint.

### 3. Config support for optional profile-scoped port mapping

The unified `proxy` config now supports:

- `proxy.port`: shared preferred default for OpenAI-compatible proxy instances
- `proxy.profile_ports`: optional per-profile port overrides

That means users can keep the current default behavior, or explicitly pin certain profiles to known ports when they need deterministic separation.

The important point is that profile-specific ports are now supported without forcing every user to configure them.

### 4. Minimal CLI behavior changes

The `ccs proxy` UX remains mostly intact for the normal case.

Single-profile usage still works as before.

The only meaningful CLI behavior adjustment is when multiple proxies are running at once:

- `ccs proxy status` can now report multiple running profile proxies
- `ccs proxy stop` can stop all proxies, or a specific profile when one is provided
- `ccs proxy activate` now asks for an explicit profile only when activation would otherwise be ambiguous

That preserves the current DX for most users while making the multi-profile case tractable instead of broken.

### 5. Ignore local-only agent override files

This PR also updates `.gitignore` to exclude local agent override files such as `AGENTS.override.md` and `CLAUDE.local.md`.

These files are useful for local assistant workflows, but they are repository-local/private guidance rather than shareable project source. Ignoring them reduces the chance of accidentally committing machine-local or user-local agent instructions.

## Why This Design

I intentionally kept this fix narrow.

I considered broader routing or dashboard-level redesigns, but the underlying problem here was lifecycle/state isolation, not request transformation or provider resolution.

This implementation was chosen because it is the smallest viable structural fix that:

- removes the singleton assumption
- supports profile-scoped port mapping
- avoids breaking the existing default path
- keeps the current CLI mental model largely unchanged
- does not require invasive changes to request routing or provider adapters

In other words: this fixes the actual architectural pressure point without over-expanding the scope.

### What gets better

- Different OpenAI-compatible settings profiles can run concurrently on separate local ports.
- Profile aliases like `ccg` and `ccgm` no longer have to collide in one shared proxy runtime.
- Users who need stable per-profile ports can configure them explicitly.
- Users who do not care about multi-profile separation keep nearly the same workflow they already had.
- Local-only agent override files are less likely to leak into commits accidentally.

### What stays the same

- The default shared proxy port remains `3456` unless overridden.
- Existing single-profile usage remains straightforward.
- Request routing and upstream model transformation behavior are unchanged.

## Testing

I validated the change at multiple levels.

### Targeted proxy tests

- Added/updated integration coverage to verify that two different OpenAI-compatible profiles can run at the same time on different ports.
- Added unit coverage for profile-scoped preferred port resolution.

### Full repository validation

Ran:

- [x] `bun run validate`
- [x] `bun run validate:ci-parity`
- [x] `cd ui && bun run validate`

Notable validation results:

- Full backend/unit/integration/npm test suite passed
- CI parity gate passed
- UI validation passed

## Checklist

## Docs Impact

Docs impact: `minor`

Action: `no update needed`

Reasoning: this changes runtime behavior and CLI help output, but it does not add a large new user-facing surface area. The main discoverability points are already reflected in `ccs proxy --help`, and the config shape is intentionally small (`proxy.port` and `proxy.profile_ports`). If maintainers want, I can follow up with a short config example in docs, but I kept this PR focused on the structural fix itself.

## Notes for Review

A few things I expect reviewers may care about:

- This PR does not change the OpenAI-compatible request/response transformation layer.
- It does not alter upstream model routing semantics.
- It does not force profile-specific ports; it only enables them.
- It explicitly favors backward-compatible default behavior for the single-profile case.
- The `.gitignore` update is intentionally small and only targets local assistant override files.

If preferred, I can split the config support and lifecycle isolation into separate PRs, but they are closely coupled enough that reviewing them together felt more coherent.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,3 +139,8 @@ jobs:
 
       - name: Test
         run: bun run test:all
+
+      - name: Test CLI e2e
+        env:
+          CCS_E2E_SKIP_BUILD: '1'
+        run: bun run test:e2e
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,9 +52,7 @@ jobs:
             ${{ runner.os }}-bun-
 
       - name: Ensure dependencies
-        run: |
-          [ -d node_modules ] || bun install --frozen-lockfile
-          [ -d ui/node_modules ] || (cd ui && bun install --frozen-lockfile)
+        run: bash scripts/ensure-deps.sh
 
       - name: Run ${{ matrix.check.name }}
         run: ${{ matrix.check.cmd }}
@@ -88,9 +86,7 @@ jobs:
             ${{ runner.os }}-bun-
 
       - name: Ensure dependencies
-        run: |
-          [ -d node_modules ] || bun install --frozen-lockfile
-          [ -d ui/node_modules ] || (cd ui && bun install --frozen-lockfile)
+        run: bash scripts/ensure-deps.sh
 
       - name: Build
         run: bun run build:all
@@ -133,9 +129,7 @@ jobs:
             ${{ runner.os }}-bun-
 
       - name: Ensure dependencies
-        run: |
-          [ -d node_modules ] || bun install --frozen-lockfile
-          [ -d ui/node_modules ] || (cd ui && bun install --frozen-lockfile)
+        run: bash scripts/ensure-deps.sh
 
       - name: Download dist artifact
         uses: actions/download-artifact@v4
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -49,6 +49,11 @@ jobs:
       - name: Validate (typecheck + lint + format + tests)
         run: bun run validate
 
+      - name: Test CLI e2e
+        env:
+          CCS_E2E_SKIP_BUILD: '1'
+        run: bun run test:e2e
+
       - name: Release
         id: release
         env:
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -41,9 +41,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
 
       - name: Install dependencies
-        run: |
-          bun install --frozen-lockfile
-          cd ui && bun install --frozen-lockfile
+        run: bash scripts/ensure-deps.sh
 
       - name: Build
         run: bun run build:all
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,6 +45,11 @@ jobs:
       - name: Validate (typecheck + lint + format + tests)
         run: bun run validate
 
+      - name: Test CLI e2e
+        env:
+          CCS_E2E_SKIP_BUILD: '1'
+        run: bun run test:e2e
+
       - name: Release
         id: release
         env:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,9 +37,7 @@ jobs:
           bun-version: '1.3.9'
 
       - name: Install dependencies
-        run: |
-          bun install --frozen-lockfile
-          cd ui && bun install --frozen-lockfile
+        run: bash scripts/ensure-deps.sh
 
       - name: Build package
         run: bun run build:all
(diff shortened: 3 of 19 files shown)


innocarpe:
After 13 review rounds, I think this PR is finally ready for human review :)

kaitranntt (approved):
Thanks for pushing this through all 13 rounds. The core design in your branch was the right fix: profile-scoped proxy lifecycle/state instead of a singleton bridge.

I did a full maintainer red-team pass before landing it. Two real stale-state edge cases showed up during review, so I pushed three small follow-up commits on top of your branch rather than replacing or squashing your work:
- `fix(proxy): harden stale daemon ownership checks`
- `test(proxy): gate proxy e2e coverage in checks`
- `docs(proxy): document multi-profile proxy lifecycle`

Those keep your original commit history intact, add regression coverage for the stale-session / stale-pid paths, and wire the CLI e2e coverage into CI and release validation.

Verification on the updated branch:
- `bun run validate`
- `bun run test:e2e`

This is in good shape to merge, and I’m preserving the full commit stack with a merge commit.

innocarpe:
@kaitranntt  Thanks so much for the thorough red-team review and for taking the time to push the follow-up fixes all the way through to merge.

Also, the AI review workflow in this repo made it much easier to align with the project’s standards and expected quality bar throughout the process, which I really appreciated.

Thanks again for the thoughtful review, the additional fixes, and for getting this over the line :)

PR #1059, opened 2026-04-21
fix(settings-profile): preserve nested model intent for agent team launches

innocarpe:
## Summary

This PR fixes the nested environment inheritance bug affecting Claude Code Team and subagent launches from CCS settings profiles.

Today, the settings-profile Claude launch path mixes two different concerns into the same inherited runtime lane:

- the parent session's model intent
- the parent session's provider/auth/routing transport

That sounds subtle, but it is exactly why Team flows break in practice.

When a user launches Claude Code through CCS with a settings profile that ultimately routes through the Codex/OpenAI-compatible bridge, nested teammates can inherit more than they should. Instead of just inheriting the parent model defaults, they can also inherit proxy transport settings like `ANTHROPIC_BASE_URL` and auth state like `ANTHROPIC_AUTH_TOKEN`.

That breaks the runtime boundary.

In practice, the child agent is no longer just "using the same model family as the parent." It is effectively being forced back onto the parent's provider lane. That is how failures like this show up:

```text
API Error: 503 auth_unavailable: no auth available (providers=codex, model=gpt-5-codex-mini)
```

This PR fixes that by preserving nested model intent while selectively stripping inherited Anthropic routing/auth environment variables for settings-profile Claude launches.

The result is the behavior users actually want:

- parent Claude sessions can still express model defaults like `gpt-5.4`
- nested Team/subagent launches can still inherit that model intent
- but they no longer inherit stale or incorrect provider transport/auth state from the parent process

스크린샷 2026-04-21 오후 3 54 15

## Problem Background

The problem was not simply "a haiku mapping bug" or "mini got chosen somewhere."

The deeper issue was boundary leakage between parent and child Claude runtimes.

CCS already had logic to aggressively inject effective settings env into top-level Claude launches. That made sense for the original goal:

- keep stale `ANTHROPIC_*` values from previous sessions from poisoning the active profile
- make the top-level Claude session faithfully reflect the current CCS profile settings

The catch is that this design was optimized for the parent launch, not for nested Team/subagent boundaries.

Once a settings-profile Claude session was up, nested launches could end up inheriting the parent's full `ANTHROPIC_*` lane, including transport/auth values. At that point the child is not making a fresh runtime decision anymore. It is getting dragged through the parent's provider tunnel.

From a user perspective, that means the Team abstraction stops feeling real. You asked for nested agents to keep the parent's model intent, but instead some of them leak into the wrong provider route and fail on spawn.

## Root Cause

The failure came from a combination of behaviors in the existing implementation:

1. `src/ccs.ts` passed effective settings env aggressively for settings-profile launches.
2. `src/utils/shell-executor.ts` merged `process.env` back into nested Claude launches.
3. Settings-profile launches did not have a selective nested-env boundary.
4. The inherited parent env could therefore carry both:
   - model defaults such as `ANTHROPIC_MODEL`, `ANTHROPIC_DEFAULT_SONNET_MODEL`, `ANTHROPIC_DEFAULT_HAIKU_MODEL`
   - routing/auth values such as `ANTHROPIC_BASE_URL`, `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_API_KEY`
5. In Codex bridge scenarios, that meant nested agents could be interpreted through the parent's provider/auth lane instead of only inheriting its model intent.

So the real bug was not "children inherit too much model state."

The real bug was "children inherit the wrong kind of state."

### 1. Added selective Anthropic routing stripping in `shell-executor`

`src/utils/shell-executor.ts` now distinguishes between two isolation modes:

- full `ANTHROPIC_*` stripping for account/default profile cases that need full routing isolation
- selective routing/auth stripping for settings-profile Claude launches that should preserve model intent

Specifically, the new selective path removes inherited:

- `ANTHROPIC_BASE_URL`
- `ANTHROPIC_AUTH_TOKEN`
- `ANTHROPIC_API_KEY`

while preserving model-selection state such as:

- `ANTHROPIC_MODEL`
- `ANTHROPIC_DEFAULT_OPUS_MODEL`
- `ANTHROPIC_DEFAULT_SONNET_MODEL`
- `ANTHROPIC_DEFAULT_HAIKU_MODEL`
- `ANTHROPIC_SMALL_FAST_MODEL`

This is the whole fix.

Keep the parent's model lane. Drop the parent's transport lane.

### 2. Updated settings-profile Claude launch wiring in `src/ccs.ts`

`src/ccs.ts` now builds a dedicated `claudeRuntimeEnvVars` payload for settings-profile Claude launches.

That payload intentionally keeps CCS-managed runtime flags but sets:

- `CCS_PROFILE_TYPE=settings`
- `CCS_STRIP_INHERITED_ANTHROPIC_ENV=1`

Then the settings-profile Claude path calls `execClaude()` with that runtime env instead of blindly passing the broader effective env set.

This means:

- non-Claude targets still receive the effective credentials they need
- Claude launches that already pass `--settings` use that settings file as the provider source of truth
- nested Team/subagent launches inherit the parent's model defaults, not its proxy tunnel

### 3. Added direct unit coverage for the selective filter policy

`tests/unit/utils/claudecode-env-stripping.test.ts` now verifies the core policy directly:

- routing/auth env is removed
- model defaults remain present

This is the low-level guardrail for the new behavior.

### 4. Added launch-path coverage through the real settings-profile wiring

`tests/unit/targets/settings-profile-browser-launch.test.ts` now verifies the higher-level path through `src/ccs.ts`:

- settings-profile Claude launch passes `CCS_STRIP_INHERITED_ANTHROPIC_ENV=1`
- inherited routing/auth env is blank in the child launch
- inherited model defaults like `ANTHROPIC_MODEL=gpt-5.4` and `ANTHROPIC_DEFAULT_SONNET_MODEL=gpt-5.4` are still preserved

That matters because this bug was not purely a helper-function bug. It lived in the interaction between `src/ccs.ts` and `src/utils/shell-executor.ts`.

## Why This Design

I kept this fix narrow on purpose.

I did not change model mapping policy. I did not try to redesign Team spawning. I did not change the Codex bridge itself.

The actual pressure point was the nested runtime boundary.

The right fix here is:

- allow model inheritance
- block provider/auth inheritance

That is the smallest structural change that matches the desired behavior.

It preserves the existing CCS mental model for parent sessions, while stopping nested child launches from getting pinned to the wrong provider/auth route.

Anything broader would have made the diff bigger without solving the core failure mode more cleanly.

### What gets better

- Nested Team/subagent launches from settings-profile Claude sessions can keep the parent's model intent.
- Child agents no longer inherit stale Codex/OpenAI-compatible routing/auth env from the parent process.
- The specific `providers=codex, model=gpt-5-codex-mini` style failure mode should be much less likely in nested launches.
- The settings file remains the source of truth for provider transport, which is where that decision belongs.

### What stays the same

- CCS still supports the existing settings-profile Claude flow.
- Parent sessions can still carry model defaults like `gpt-5.4`, `gpt-5.4-mini`, and related tier mappings.
- Account/default profile launch isolation behavior remains intact.
- This PR does not change upstream request transformation, provider adapters, or model alias normalization.

## Testing

I validated the change at two layers.

### Targeted unit coverage

Ran:

- [x] `bun test tests/unit/utils/claudecode-env-stripping.test.ts`
- [x] `bun test tests/unit/targets/settings-profile-browser-launch.test.ts`

Results:

- `tests/unit/utils/claudecode-env-stripping.test.ts` → `14 pass, 0 fail`
- `tests/unit/targets/settings-profile-browser-launch.test.ts` → `7 pass, 0 fail`

### Push gate / build verification

The feature branch push also passed the repo's fast pre-push verification path, including:

- `tsc --noEmit`
- `eslint src/ --fix`
- `prettier --check src/`
- `bun run ui:build && bun run build:server`

That gave a clean build on the merged branch state before PR creation.

## Checklist

## Docs Impact

Docs impact: `minor`

Action: `no update needed`

Reasoning: this PR changes nested runtime behavior and regression coverage, but it does not add a new end-user command surface or config shape. The main value here is correctness at the Team/subagent boundary, not a new user-facing workflow. If maintainers want, a follow-up note in docs about nested model inheritance vs routing inheritance could make sense, but I kept this PR scoped to the bug fix itself.

## Notes for Review

A few things I expect reviewers may care about:

- This PR intentionally preserves nested model inheritance. It does not remove it.
- The selective filter targets routing/auth keys, not model defaults.
- The bug fix spans both the helper layer and the settings-profile launch wiring. That is why there are two tests at different levels.
- This PR does not claim to fully reproduce every Team runtime path in integration. It hardens the exact boundary that was leaking.
- The uncommitted investigation note under `docs/reports/` is intentionally not part of this PR.

### DIFF
--- a/src/ccs.ts
+++ b/src/ccs.ts
@@ -1359,8 +1359,20 @@ async function main(): Promise<void> {
         console.error(info(`Global env: ${envNames}`));
       }
 
-      // Explicitly inject effective settings env vars so stale ANTHROPIC_*
-      // values from prior sessions cannot leak into the active profile.
+      // For Claude target launches that already pass `--settings`, keep runtime
+      // env limited to CCS-managed flags so nested Team/subagent sessions do not
+      // inherit profile-specific ANTHROPIC_* routing from the parent process.
+      const claudeRuntimeEnvVars: NodeJS.ProcessEnv = {
+        ...globalEnv,
+        ...(inheritedClaudeConfigDir ? { CLAUDE_CONFIG_DIR: inheritedClaudeConfigDir } : {}),
+        ...webSearchEnv,
+        ...imageAnalysisEnv,
+        ...(browserRuntimeEnv || {}),
+        CCS_PROFILE_TYPE: 'settings',
+        CCS_STRIP_INHERITED_ANTHROPIC_ENV: '1',
+      };
+
+      // Non-Claude targets still need effective credentials injected directly.
       const envVars: NodeJS.ProcessEnv = {
         ...globalEnv,
         ...settingsEnv,
@@ -1472,7 +1484,7 @@ async function main(): Promise<void> {
         settingsPath: expandedSettingsPath,
       });
 
-      execClaude(claudeCli, launchArgs, { ...envVars, ...traceEnv });
+      execClaude(claudeCli, launchArgs, { ...claudeRuntimeEnvVars, ...traceEnv });
     } else if (profileInfo.type === 'account') {
       // NEW FLOW: Account-based profile (work, personal)
       // All platforms: Use instance isolation with CLAUDE_CONFIG_DIR
--- a/src/ccs.ts
+++ b/src/ccs.ts
@@ -80,7 +80,7 @@ import { handleError, runCleanup } from './errors';
 import { tryHandleRootCommand } from './commands/root-command-router';
 
 // Import extracted utility functions
-import { execClaude } from './utils/shell-executor';
+import { execClaude, stripAnthropicRoutingEnv } from './utils/shell-executor';
 import { isDeprecatedGlmtProfileName, normalizeDeprecatedGlmtEnv } from './utils/glmt-deprecation';
 import { maybeWarnAboutResumeLaneMismatch } from './auth/resume-lane-warning';
 import { createLogger } from './services/logging';
@@ -1360,10 +1360,11 @@ async function main(): Promise<void> {
       }
 
       // For Claude target launches that already pass `--settings`, keep runtime
-      // env limited to CCS-managed flags so nested Team/subagent sessions do not
-      // inherit profile-specific ANTHROPIC_* routing from the parent process.
+      // env free of ANTHROPIC routing/auth while preserving non-routing profile
+      // env so nested Team/subagent sessions can still inherit model intent and
+      // other profile-scoped runtime flags.
       const claudeRuntimeEnvVars: NodeJS.ProcessEnv = {
-        ...globalEnv,
+        ...stripAnthropicRoutingEnv({ ...globalEnv, ...settingsEnv }),
         ...(inheritedClaudeConfigDir ? { CLAUDE_CONFIG_DIR: inheritedClaudeConfigDir } : {}),
         ...webSearchEnv,
         ...imageAnalysisEnv,
--- a/src/delegation/headless-executor.ts
+++ b/src/delegation/headless-executor.ts
@@ -16,8 +16,13 @@ import { type ExecutionOptions, type ExecutionResult, type StreamMessage } from 
 import { StreamBuffer, formatToolVerbose } from './executor/stream-parser';
 import { buildExecutionResult } from './executor/result-aggregator';
 import { getCcsDir, getModelDisplayName, loadSettings } from '../utils/config-manager';
+import { getGlobalEnvConfig } from '../config/unified-config-loader';
 import { getProfileLookupCandidates } from '../utils/profile-compat';
-import { getClaudeLaunchEnvOverrides, stripClaudeCodeEnv } from '../utils/shell-executor';
+import {
+  getClaudeLaunchEnvOverrides,
+  stripAnthropicRoutingEnv,
+  stripClaudeCodeEnv,
+} from '../utils/shell-executor';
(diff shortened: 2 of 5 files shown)


innocarpe:
@kaitranntt Heads up: the failing `typecheck` check does not appear to be caused by the
changes in this PR.

Both failures happen in `scripts/ensure-deps.sh` during dependency sync, before TypeScript checking starts:

- Run `24710972536`: root `bun install` failed with `FileNotFound: failed copying files from cache to destination for package @typescript-eslint/scope-manager`
- Run `24711940055`: UI `bun install` failed with `FileNotFound: failed copying files from cache to destination for package date-fns`

This branch only touches:

- `src/ccs.ts`
- `src/utils/shell-executor.ts`
- `tests/unit/targets/settings-profile-browser-launch.test.ts`
- `tests/unit/utils/claudecode-env-stripping.test.ts`

There are no changes to dependency manifests or lockfiles in either the root project or `ui/`.

Since the failing package changed between runs and the rest of the checks are otherwise behaving as expected, this looks more like a Bun install/cache flake in CI than a PR-specific dependency regression.

Could you take a look at whether `scripts/ensure-deps.sh` should be hardened a bit, for example with retry or cache recovery around `bun install`?

kaitranntt:
@innocarpe Much appreciated for the huge contribution here, especially the detailed analysis and root-cause write-up. That level of context made review and merge much easier. Please keep this level of interaction on future PRs too.

PR #1104, opened 2026-04-26
test(cliproxy): replace process-wide mock.module with DI seams

sgaluza:
## Summary

Fixes a latent CI flake: 6 tests in `cliproxy-stats-routes-*` fail when run after either `service-manager-startup.test.ts` or `version-checker-stale-cache.test.ts`. Root cause is bun's `mock.module()` being process-wide and **not** undone by `mock.restore()`.

The bug is masked today by file-ordering luck. I hit it locally on a clean `dev` checkout via `bun run test:fast` (6 fail). Adding/reordering tests anywhere in the suite can flip the ordering and surface this on PRs unrelated to cliproxy.

## Root cause

bun's `mock.module()` patches modules at the process level. `mock.restore()` does not undo it. Two test files were applying stubs at the top level:

- `tests/unit/cliproxy/service-manager-startup.test.ts` — 8 `mock.module()` calls patching `binary-manager`, `proxy-detector`, `startup-lock`, `config-generator`, `stats-fetcher`, `session-tracker`, `auth/token-refresh-config`, `auth/token-refresh-worker`.
- `tests/unit/cliproxy/version-checker-stale-cache.test.ts` — one test that patched `version-checker.checkForUpdates` via `mock.module()` after a prior `ensureBinary` call.

Subsequent files that transitively imported any of these (notably `cliproxy-stats-routes-model-update.test.ts`) inherited the leaked stubs and asserted against fixture data that never materialized.

## Fix: dependency-injection seams

Surface-minimal — every new param is **optional** and defaulted to the real implementation. No production behavior change.

**`src/cliproxy/types.ts`** — added optional `checkForUpdatesFn?` to `BinaryManagerConfig`. Lifecycle reads `config.checkForUpdatesFn ?? checkForUpdates`.

**`src/cliproxy/service-manager.ts`** — added `EnsureCliproxyServiceDeps` interface with 4 optional fields (`ensureBinaryFn`, `detectRunningProxyFn`, `configNeedsRegenerationFn`, `withStartupLockFn`), defaulted to the real implementations.

**Tests** rewritten to inject stubs via the new seams, replacing 9 top-level `mock.module()` calls with plain function arguments. `mock.restore()` is no longer needed because nothing global is mutated.

Diff: `+74 / -86` across 5 files.

## Verification

- Before fix on clean `dev`: `bun run test:fast` → **6 fail**
- After fix: `bun run test:fast` → **2540 pass / 0 fail / 1 skip**
- `tsc --noEmit`, `eslint src/`, `prettier --check src/` — all green
- Pre-push gate (typecheck + lint + format:check + test:fast) passes

## Test plan

- [ ] CI green on this PR
- [ ] Production code paths still call real `checkForUpdates`, `ensureCLIProxyBinary`, `detectRunningProxy`, `configNeedsRegeneration`, `withStartupLock` (defaults preserved)
- [ ] `cliproxy-stats-routes-*` no longer fails when run after `service-manager-startup` or `version-checker-stale-cache`

Built [OnSteroids](https://onsteroids.ai)

### DIFF
--- a/src/cliproxy/binary/lifecycle.ts
+++ b/src/cliproxy/binary/lifecycle.ts
@@ -57,7 +57,8 @@ function clampToMaxStable(version: string | undefined, verbose: boolean): string
 async function handleAutoUpdate(config: BinaryManagerConfig, verbose: boolean): Promise<void> {
   const backend: CLIProxyBackend = config.backend ?? DEFAULT_BACKEND;
   const backendLabel = getBackendLabel(backend);
-  const updateResult = await checkForUpdates(config.binPath, config.version, verbose, backend);
+  const checkFn = config.checkForUpdatesFn ?? checkForUpdates;
+  const updateResult = await checkFn(config.binPath, config.version, verbose, backend);
   const currentVersion = updateResult.currentVersion;
   const latestVersion = updateResult.latestVersion;
 
--- a/src/cliproxy/service-manager.ts
+++ b/src/cliproxy/service-manager.ts
@@ -139,6 +139,18 @@ export interface ServiceStartResult {
 }
 
 /**
+ * Test-only seams for ensureCliproxyService. Production callers omit this and
+ * get the real implementations. Tests inject stubs to avoid bun's
+ * `mock.module()`, which is process-wide and leaks across test files.
+ */
+export interface EnsureCliproxyServiceDeps {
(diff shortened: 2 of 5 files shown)


kaitranntt:
@sgaluza thanks for the fix and the clear root-cause write-up. Much appreciated.

PR #1356, opened 2026-05-24
fix(codex): enforce local cliproxy base URL for ccsxp repair

kaitranntt:
### Motivation
- Restore the previous local-endpoint safety invariant so ccsxp does not preserve arbitrary remote `model_providers.cliproxy.base_url` values that could exfiltrate an injected CLIProxy API key and request data.
- Prevent attacker-controlled or remote endpoints from being treated as "ready" and left in place by the Codex provider repair logic.

### Description
- Tighten `isValidCodexCliproxyBaseUrl` in `src/targets/codex-cliproxy-provider-config.ts` to only accept loopback hosts (`localhost`, `127.0.0.1`, `::1`) with `http:` and the exact `/api/provider/codex` path.
- Keep provider-repair behavior but force fallback to the managed local base URL whenever the configured `base_url` is remote or otherwise unsafe by using the stricter validation in `resolveProviderBaseUrl`/`isProviderReady`.
- Update unit tests in `tests/unit/targets/codex-cliproxy-provider-config.test.ts` to assert that non-loopback/remote URLs are repaired back to the managed local cliproxy URL while preserving a custom `env_key`.

### Testing
- Ran `bun test tests/unit/targets/codex-cliproxy-provider-config.test.ts`, and the test suite for the modified code passed (all tests in that file passed).
- `tsc --noEmit` was executed as part of validation and succeeded for the changed files.
- The repository `format:check`/Prettier pre-commit step flagged unrelated formatting issues in other files, so the formatting hook prevented a normal commit for unrelated files; the security fix and its tests were verified independently and included in this PR.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a11fa79d0bc8330b91db52e3d8df0b5)

### DIFF
--- a/src/targets/codex-cliproxy-provider-config.ts
+++ b/src/targets/codex-cliproxy-provider-config.ts
@@ -56,7 +56,10 @@ function isValidCodexCliproxyBaseUrl(value: unknown): value is string {
   if (!trimmed) return false;
   try {
     const url = new URL(trimmed);
-    return url.protocol === 'http:' || url.protocol === 'https:';
+    const hostname = url.hostname.toLowerCase();
+    const isLoopbackHost =
+      hostname === 'localhost' || hostname === '127.0.0.1' || hostname === '::1';
+    return isLoopbackHost && url.protocol === 'http:' && url.pathname === '/api/provider/codex';
   } catch {
     return false;
   }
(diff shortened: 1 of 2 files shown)


kaitranntt:
Closing: review found this fix is too restrictive — forces base_url to loopback + `/api/provider/codex` only, which breaks the documented remote cliproxy use case (e.g. `192.168.2.84:8317` per ccs-cliproxy-architecture.md). Will regenerate with a relaxed validator that allows configured remote hosts via allowlist instead of forcing loopback.

PR #1514, opened 2026-06-11
feat(cliproxy): account pools with auto-continue at usage limits

### LINKED ISSUE #1464: feat: Auto-continue on next Claude Code account when limit is reached
I use CCS with 3 Claude Code accounts. For large tasks, it would be great if CCS could automatically continue the same task on the next available account/session when the current one hits a usage limit.

Suggested flow:

- detect limit reached
- switch to next configured account
- pass current context / summary
- continue the task automatically

This could be an optional setting/skill, e.g. autoContinueOnLimit: true.

Would make long tasks much smoother without manual switching.

kaitranntt:
Closes #1464

## What

Users with multiple Claude subscription accounts can pool them behind `ccs claude`. When the active account hits its usage limit, the proxy puts it on cooldown and the session continues on the next account automatically. The same applies to other CLIProxy providers with 2 or more accounts.

Setup for the issue #1464 scenario is three commands and zero config edits:

```
ccs claude --auth
ccs claude --auth --add   (repeat per account, pool prompt appears at the second)
ccs claude
```

## How it works

- Pool opt in enables quota cooling (`disable-cooling: false`), fill-first routing with session affinity, and a retry cap (`max-retry-credentials: 3`). Non-pool users keep their current generated config unchanged.
- Fill-first plus affinity keeps prompt caches warm: a session stays on one account until it is actually exhausted, then hops once and sticks to the next account. Per-request rotation would burn several times the quota on long sessions, so it is never selected.
- Drain order is managed through a top-level `priority` field in auth files, editable with `ccs cliproxy accounts order <provider>`. Writes go through the management API when the proxy is running so they are not clobbered.
- `ccs cliproxy quota` shows pool context: drain order, per-account state (available, cooling with reset time, paused), strategy and affinity. The dashboard routing card reflects pool mode.
- Existing users with 2 or more native Claude profiles get a one-time, dismissible hint in `ccs doctor`, at profile launch, and in `ccs auth create`.
- Rollback: `ccs cliproxy pool disable` restores the previous routing config cleanly.

## Merged phase PRs

- #1509 claude pool gap closure: model-neutral launch, shadow warning, provider ban copy
- #1510 pool routing defaults and safety rails
- #1511 managed drain order
- #1512 onboarding hints for existing multi-profile users
- #1513 pool visibility in quota output and dashboard routing card

## Notes

- Docs: kaitranntt/ccs-docs#57 (draft, `/features/pool`) ships alongside the release.
- The temporary CI trigger entry for this epic branch was removed before promotion; PR CI triggers are back to `[main, dev]`.
- Live multi-account verification (limit wire format, hop latency, mid-stream behavior) is tracked as the plan's Phase 1 spike and will finalize the docs caveat numbers.


kaitranntt:
Follow up on the two focus areas from the updated review:

**Legacy Notice**: valid, fixed in cff9008f. The legacy-config skip in the pool opt-in prompt is now silent. Legacy installs get pool guidance only from ccs doctor, which is the once-per-install surface the phase design assigned to that population. Printing on every account add was a nag with no dismissal available.

**False Positive**: refuted by the profile-registry schema. Every profiles.json entry with type 'account' is an isolated Claude instance (CLAUDE_CONFIG_DIR lane); there is no CLI discriminator because Gemini and Codex multi-account live in CLIProxy auth files, never in profiles.json. The counter cannot over-count. Added a comment on countNativeClaudeProfiles in the same commit so this is clear at the call site.

kaitranntt:
Round 2 verdicts, addressed in 7b8a6f5c:

**Premature hint**: confirmed and fixed. The create-command hint now fires only inside the exit-code-0 branch, after the profile exists. The old pre-create placement burned the once-per-install dismissal whenever creation failed or rolled back.

**Custom path stripping**: the literal claim is off, the quoted filter runs only in Priority 2 (the default settings file); Priority 1 custom paths were already untouched. But there was a real adjacent gap: an explicit 'ccs claude --config' pick lands in that same default file, and the remote filter stripped it when the pick equaled a historical default. Fixed by gating the filter on the migration marker: once the file has been cleaned, any remaining pin is user intentional and survives. New test covers it.

**Legacy mutation**: refuted. loadOrCreateUnifiedConfig is read-only; 'create' returns an in-memory default object and never writes config.yaml (createEmptyUnifiedConfig is a pure factory), and dismissOnboardingHint independently guards legacy installs. Added a docblock so the name stops reading like a disk write.

PR #1524, opened 2026-06-15
Fix/upstream undici timeouts

kcfang:
## Summary

- undici defaults `headersTimeout`/`bodyTimeout` to 300s on every bare `Agent()` / the global dispatcher. The OpenAI-compat messages route arms an `AbortController` with `REQUEST_TIMEOUT_MS` (600s, tunable via `CCS_OPENAI_PROXY_REQUEST_TIMEOUT_MS`), but undici's 300s per-phase timeout fires first — so any upstream that stays header/body-silent past ~5 min (self-hosted LLMs with long queue/prefill phases) has its socket torn down. For streaming requests this surfaces downstream as `The socket connection was closed unexpectedly` rather than the proxy's timeout response, and the configured request timeout + its env override are effectively dead code beyond 300s.
- **Production evidence** (llama.cpp/llama-swap behind ccs, ~16-way agent fan-out): every request that stayed silent past the 300s floor died; the longest survivor in the gateway access log was 4m19s (259s), right under the floor.
- **Fix:** derive `headersTimeout`/`bodyTimeout` from `getRequestTimeoutMs()` + a 30s grace margin (so the `AbortController` stays the single authority) and apply them to every dispatcher used for upstream fetches:
  - `messages-route.ts`: new `buildUpstreamAgentTimeouts()`; the secure path now passes an explicit dispatcher (lazily built via `createGlobalFetchProxyDispatcher(timeouts)`, preserving `HTTP(S)_PROXY` routing, falling back to a plain `Agent`); the per-request insecure dispatcher carries the same options.
  - `proxy-server.ts`: the shared insecure dispatcher carries the same options.
  - `fetch-proxy-setup.ts`: `RoutingProxyDispatcher` / `createGlobalFetchProxyDispatcher` accept optional agent timeout options (default `{}` → no behavior change for the existing global setup call).
- No behavior change for requests faster than 300s, or for the global fetch proxy setup defaults. The `insecureTls` log field now reflects actual insecure-TLS usage (it previously keyed off `dispatcher !== undefined`, which would otherwise become always-true once the secure path also gets a dispatcher).

## Testing

- [x] `bun run validate:ci-parity` before requesting review — green except the pre-existing `ccs-browser MCP server … becomes stale` test, which fails identically on untouched `dev` (Developer-Mode symlink / environment issue, unrelated to this change)
- [x] `bun run test:e2e` —  35 pass, 0 fail, 105 expect() calls
- [x] `cd ui && bun run validate`

New/updated tests in `tests/unit/proxy/messages-route.test.ts`:
- `buildUpstreamAgentTimeouts` keeps both timeouts above the default and the env-overridden request timeout.
- Secure-path regression: the upstream fetch must receive an explicit dispatcher (previously `undefined` → undici defaults).
- Existing insecure-dispatcher tests unchanged and green.

## Checklist

- [x] Base branch is `dev`
- [x] Branch name follows `fix/*`
- [ ] Relevant `--help` output updated if CLI behavior changed — N/A (no CLI surface change)
- [x] README or local docs updated if user-facing behavior changed — documented `CCS_OPENAI_PROXY_REQUEST_TIMEOUT_MS` in `docs/openai-compatible-providers.md`

## Docs Impact

Docs impact: `minor`

Action: added a Troubleshooting entry in `docs/openai-compatible-providers.md` for the "socket connection was closed unexpectedly" symptom on slow upstreams, pointing at `CCS_OPENAI_PROXY_REQUEST_TIMEOUT_MS` (previously undocumented).

### DIFF
--- a/docs/openai-compatible-providers.md
+++ b/docs/openai-compatible-providers.md
@@ -327,6 +327,14 @@ That flag is respected by both:
 - CCS now preserves upstream rate-limit errors and retry headers
 - Empty or malformed provider JSON is returned as Anthropic-style `api_error`
 
+### Slow upstreams: `socket connection was closed unexpectedly`
+
+- Long-running upstreams (self-hosted LLMs with queue/prefill phases) can stay
+  silent for minutes before the first or next token. The proxy already allows up
+  to 10 minutes per request; set `CCS_OPENAI_PROXY_REQUEST_TIMEOUT_MS` in the
+  profile settings to raise or lower that ceiling.
+- Restart the proxy after changing the setting.
+
 ### Requests route to the wrong model/profile
 
 - Use an explicit selector such as `profile:model`
--- a/src/proxy/server/messages-route.ts
+++ b/src/proxy/server/messages-route.ts
@@ -10,9 +10,16 @@ import {
 import { ProxySseStreamTransformer } from '../transformers/sse-stream-transformer';
 import { isAnthropicPassthroughProfile, resolveOpenAIChatCompletionsUrl } from '../upstream-url';
 import { createLogger } from '../../services/logging';
+import {
+  createGlobalFetchProxyDispatcher,
+  type UpstreamAgentTimeoutOptions,
+} from '../../utils/fetch-proxy-setup';
 import { pipeWebResponseToNode, readRawBody, writeJson } from './http-helpers';
 
 const REQUEST_TIMEOUT_MS = 600_000;
+// Keep undici's per-phase timeouts above the explicit request timeout so the
+// AbortController is the single authority on when an upstream request dies.
+const UPSTREAM_TIMEOUT_GRACE_MS = 30_000;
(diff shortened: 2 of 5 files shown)


kaitranntt (approved):
Reviewed and verified against the pinned `undici@5.29.0`:

- `ProxyAgent({ uri, ...agentOptions })` is the correct v5 options-object signature, so the proxy-routing path keeps the new `headersTimeout`/`bodyTimeout` while honoring `HTTP(S)_PROXY`.
- Timeout model is sound: per-phase undici timeouts are set to `request_timeout + 30s` so the `AbortController` stays the single authority on when an upstream dies, instead of undici's silent 300s cutoff closing slow self-hosted streams.
- The memoized default dispatcher reads `CCS_OPENAI_PROXY_REQUEST_TIMEOUT_MS` once at proxy start, so a change requires a proxy restart — correctly documented in the docs update.
- Tests cover the ceiling math, the env-override tracking, and the secure-path regression (explicit dispatcher rather than undici defaults).

CI Gate is green and the change carries no public-contract change. Looks good to merge into `dev`.

Issue #1406, opened 2026-05-29
bug: detail\":\"System messages are not allowed\


matheus051004:
### Affected area

None

### What broke?

=== REQUEST INFO ===
Version: 7.1.19
URL: /api/provider/codex/v1/messages?beta=true
Method: POST
Downstream Transport: http
Upstream Transport: http
Timestamp: 2026-05-29T09:00:41.178601152-03:00

Model: 5.4, 5.3-codex and more...

Response:
{
  "type": "error",
  "error": {
    "type": "invalid_request_error",
    "message": "{\"detail\":\"System messages are not allowed\"}"
  }
}

### CCS version

8.1.0

### Operating system

Linux

### Environment details

Node.js - Linux ZorinOS/Ubuntu

### Logs, screenshots, or terminal output

```shell

```

### Before submitting

- [x] I searched existing issues first.
- [ ] I removed secrets and private data from logs/screenshots.
- [ ] I can still reproduce this on the latest released or dev build.

kaitranntt:
Fixed in PR #1407 by folding Codex-targeted system messages into the first user message before forwarding to Codex, while leaving non-Codex providers unchanged.

Verification:
- Regression coverage for Codex system-message normalization passed.
- `bun run validate` passed.
- `bun run validate:ci-parity` passed.
- PR #1407 CI passed and merged to `dev`.
- Released in `v8.1.1-dev.1`; current dev line also includes it via `v8.1.2-dev.1`.

Closing as fixed. Install with `ccs update --dev`.

KosmoCHE:
I have update to dev, but it's still unavailable for gpt-5.5-fast.

```
 API Error: 502 unknown provider for model gpt-5.5-fast. This is a server-side issue, usually temporary — try again in a moment. If it persists, check your inference gateway (127.0.0.1:8317).
```

kaitranntt:
Reopening — the `gpt-5.5-fast` 502 reported above is a **distinct, still-unresolved defect** from the "System messages are not allowed" case that #1407 fixed. The fast-alias request never reaches the layer #1407 patched, so the fix could not have addressed it.

## Observed symptom
```
API Error: 502 unknown provider for model gpt-5.5-fast ... check your inference gateway (127.0.0.1:8317)
```

## Where the 502 actually comes from
[...]

PR #1538, opened 2026-06-16
fix: preserve Codex system prompt structure

kaitranntt:
### Motivation
- Fix a security/regression where Codex-routed requests folded privileged `system`/developer instructions into ordinary user content, weakening instruction boundaries and increasing prompt-injection risk. 
- Keep the intended Codex model tuning alias normalization while avoiding any rewrite that demotes or removes `system` messages. 

### Description
- Remove the Codex-specific system-message folding code and helper functions so top-level `system` fields and `role: 'system'` messages are forwarded unchanged. 
- Stop invoking the folding rewrite for Codex requests and only apply `applyCodexModelTuningAlias` for model tuning alias normalization. 
- Update the integration test to assert that Codex requests preserve `system` fields and `role: 'system'` messages instead of folding them into user content. 
- Apply small formatting adjustments where CI-parity Prettier drift was detected. 

### Testing
- Ran code style and linting: `bun run format` and `bun run lint:fix`, both completed and formatting drift was fixed. 
- Ran the targeted integration test `bun test ./src/cliproxy/proxy/__tests__/tool-sanitization-proxy-integration.test.ts -t "preserves Codex system"`, which passed. 
- Ran broader checks `bun run validate` and `bun run validate:ci-parity`; these exercised the full test matrix and surface unrelated pre-existing failures in other suites (not caused by this change), notably `web-server account-routes context normalization > rolls back inferred shared resource metadata when instance reconciliation fails` and `browser status > bootstraps the managed default browser profile dir before reporting attach readiness`, plus several settings-profile / browser / image-analysis / websearch launch tests in the CI-parity bucket.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a28dde7d5ec83308e5d169bd1940225)

### DIFF
--- a/src/cliproxy/proxy/__tests__/tool-sanitization-proxy-integration.test.ts
+++ b/src/cliproxy/proxy/__tests__/tool-sanitization-proxy-integration.test.ts
@@ -487,7 +487,7 @@ describe('ToolSanitizationProxy Integration', () => {
       }
     });
 
-    it('folds Codex system messages into the first user message before forwarding', async () => {
+    it('preserves Codex system fields and system-role messages before forwarding', async () => {
       const proxy = new ToolSanitizationProxy({
         upstreamBaseUrl: `http://127.0.0.1:${mockUpstreamPort}`,
       });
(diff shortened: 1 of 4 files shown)


kaitranntt:
This cannot merge as-is. The diff removes `foldCodexSystemMessages`, which was added in #1407 to fix #1406: the Codex upstream (`/api/provider/codex/v1/messages`) rejects any request carrying a top-level `system` field or a `role:"system"` message with HTTP 400 `{"detail":"System messages are not allowed"}`. Claude Code always sends a system prompt, so removing the fold re-opens #1406 and breaks the Codex happy path broadly, not just an edge case.

The rewritten integration test asserts system content is forwarded unchanged, so it encodes the regressing behavior — its green status does not validate correctness. The "preserve structure / reduce prompt-injection" framing is inverted: folding system content into the first user message is the mandated client-side upstream-compatibility shim, not a security hole. Any boundary hardening belongs upstream, not by deleting the shim.

Closing to keep #1406 fixed. If the Codex upstream now accepts system messages, reopen with a live request proving it and we can drop the fold safely.

PR #1556, opened 2026-06-16
fix: ignore malformed bar analytics date keys

kaitranntt:
### Motivation

- The merged daily/hourly analytics path could throw `RangeError: Invalid Date` and return HTTP 500 when aggregated date keys derived from non-ISO but `Date.parse`-accepted timestamps produced malformed day keys.
- The change makes the analytics path robust to malformed aggregate keys so a bad cached usage record cannot take `/api/bar/analytics` down while preserving existing analytics semantics.

### Description

- Make `dateFromDayKey` strict by validating `YYYY-MM-DD` with a regex and returning `null` for malformed or out-of-range keys instead of producing an invalid `Date` object.
- Guard daily and hourly processing in `computeBarAnalyticsFromDaily` by converting keys via `dateFromDayKey` into `activityDate` and skipping entries with invalid day keys before computing `dayDelta`, touching activity, or filling sparkline buckets.
- Compute `lastActivityAt`/`daysSinceLastActivity` from a validated `lastActivityDate` so `toISOString()` is never called on an invalid `Date`.
- Add a unit test that verifies `computeBarAnalyticsFromDaily` ignores `Date.parse`-accepted but non-`YYYY-MM-DD` aggregate keys and preserves valid analytics data; run formatting which adjusted a couple of import-line wraps in unrelated files.

### Testing

- Ran the focused unit suite `bun test tests/unit/web-server/bar-analytics.test.ts` and all tests (15) passed.
- Ran `bun run typecheck`, `bun run lint`, and `bun run format:check` and they passed after formatting was applied.
- Attempted full `bun run validate`; the typecheck/lint/format stages completed, but the full validation run was interrupted due to an unrelated/hung test in the fast test bucket, so the broader integration test bucket was not completed (earlier validation stages succeeded).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a313f299ba88330b76b27bafddf20cb)

### DIFF
--- a/src/cliproxy/quota/quota-manager.ts
+++ b/src/cliproxy/quota/quota-manager.ts
@@ -514,9 +514,8 @@ export async function reconcileExhaustedRotationAccounts(
     return [];
   }
 
-  const { pauseAccountForQuotaCooldown, restoreExpiredQuotaPauses } = await import(
-    '../accounts/account-safety'
-  );
+  const { pauseAccountForQuotaCooldown, restoreExpiredQuotaPauses } =
+    await import('../accounts/account-safety');
   restoreExpiredQuotaPauses();
 
   const config = loadOrCreateUnifiedConfig();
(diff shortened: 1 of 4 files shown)


kaitranntt:
Verified base branch is `dev`.

Remaining blocker is format-only: Prettier reports issues in:
- `src/cliproxy/quota/quota-manager.ts`
- `src/management/checks/image-analysis-check.ts`

Run `bun run format` or Prettier on those files, then the CI Gate should clear if checks stay otherwise green.

PR #1575, opened 2026-06-20
Fix CLIProxy provider routes for original backend

walker1211:
## Summary
- Route local provider profiles through the CLIProxy root endpoint when the configured backend is `original`.
- Preserve `/api/provider/<provider>` scoped routes for the `plus` backend.
- Repair stale managed Codex/Claude settings that still point local original-backend profiles at provider-scoped URLs.
- Keep API bridge metadata and Codex reasoning proxy URL reporting consistent with backend-specific routing.
- Use a type-only import for platform detector type symbols so CI-parity UI builds pass with `verbatimModuleSyntax`.

## Test plan
- [x] bun run validate
- [x] bun run validate:ci-parity
- [x] pre-push fast gate (`tsc --noEmit`, `eslint src/`, `prettier --check src/`, fast test bucket)

### DIFF
--- a/src/api/services/__tests__/cliproxy-profile-bridge.test.ts
+++ b/src/api/services/__tests__/cliproxy-profile-bridge.test.ts
@@ -1,14 +1,13 @@
 /**
- * Regression tests for the `ccs api create --cliproxy-provider claude` bridge path.
+ * Regression tests for the `ccs api create --cliproxy-provider <provider>` bridge path.
  *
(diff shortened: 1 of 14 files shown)


kaitranntt (approved):
Validated locally on the PR head: `bun run validate` + full `test:all` (387 fast + 80 slow test files) all pass, including the cliproxy-profile-bridge tests. Routing logic verified against upstream CLIProxyAPI behavior (root endpoint for original backend, scoped /api/provider for plus). LGTM.

PR #1596, opened 2026-06-24
feat(bar): per-profile subscription quota for ccs auth + ccsx auth accounts

### LINKED ISSUE #1595: feat: CCS Bar — show per-profile subscription quota for ccs auth + ccsx auth accounts
## Description

The CCS Bar shows quota only for the global default Claude/Codex login. Subscription
accounts created via `ccs auth` (isolated Claude profiles) and `ccsx auth` (isolated
Codex profiles) are not surfaced, so active subscription profiles are invisible or
unattributed in the Bar.

## Goal

Enumerate every `ccs auth` (Claude) and `ccsx auth` (Codex) profile, plus the bare
default Codex login, and show each profile's available quota in the Bar, clearly
labeled by surface (`ccs` vs `ccsx`) and profile name. Same-name profiles across
surfaces must remain distinct.

## Scope

- Server: enumerate profile registries, reuse existing quota fetchers + cache +
  circuit breaker, emit one summary row per profile with `surface`, `profile`, and an
  explicit subscription flag (pool Codex accounts also report provider `codex`, so the
  native-vs-pool split must key on the flag, not the provider string).
- Active profiles polled live; parked/expired profiles shown dimmed and not polled.
- Bar UI: group subscription rows by provider; render each provider as a horizontal
  swipe carousel with page indicators; add a surface tag chip and a dimmed parked state.

## Acceptance

- Each profile appears, labeled `provider · profile` with a `ccs`/`ccsx` tag.
- Active profiles show live quota windows; parked profiles render dimmed, never polled.
- Per-provider horizontal swipe works in the menu bar popover.
- No regression to pool accounts, spend strip, or analytics.

kaitranntt:
## Summary

Adds per-profile subscription quota to the CCS Bar. The Bar previously showed only the global default Claude/Codex login; it now enumerates every `ccs auth` (Claude) and `ccsx auth` (Codex) profile plus the bare `ccsx`/`ccs` default login, and shows each one's quota labeled by surface and profile.

Subscriptions are grouped by provider; each provider renders as a horizontal swipe carousel of its profiles with page dots. The default account leads, shown as the base command (`ccsx`/`ccs`) with a `default` badge; named profiles show their name (`ck`) plus the owning surface tag. Active profiles show live quota windows; profiles without resolvable on-disk credentials are dimmed with a `reauth` badge.

Closes #1595

## What changed

| Area | Change |
|------|--------|
| `src/web-server/usage/native-quota-collector.ts` | Per-profile enumeration + per-profile state maps; file-only Claude reader for isolated profiles (no Keychain); Codex `auth.json` reader; bare default account named `default`, read through the single global credential lookup |
| `src/web-server/routes/bar-routes.ts` | New wire fields `surface`, `profile`, `is_subscription` |
| `macos-bar/Sources/CCSBarCore/BarSummary.swift` | Decode the three new fields (backward compatible) |
| `macos-bar/Sources/CCSBarCore/BarFormatting.swift` | `is_subscription` partition; default-vs-profile title/tag helpers |
| `macos-bar/Sources/CCSBarApp/BarMenuView.swift` | Per-provider profile carousel (macOS 14 scroll paging), default account leads, tightened spacing |
| `macos-bar/Sources/CCSBarApp/BarSubscriptionCard.swift` | Surface/default tag chip + parked dimming |
| `macos-bar/Package.swift`, `macos-bar/VERSION` | Deployment target raised to macOS 14; Bar version 1.9.0 |

## Design notes

- **No Keychain fan-out.** Isolated `ccs auth` profiles read credentials from disk only; a profile without a credentials file renders as needs-reauth rather than triggering a Keychain prompt. The bare default account uses the single global credential lookup the shipped Bar already performs (one item), not a per-profile Keychain scan.
- **Account safety.** Only profiles with a valid on-disk/global token are live-polled; the existing 10-minute TTL cache, per-profile circuit breaker, and 2.5s `/summary` deadline are preserved.
- **macOS 14.** The native horizontal scroll paging (`scrollTargetBehavior(.paging)`) requires macOS 14, so the app deployment target moves from 13 to 14.

## Validation

- [x] `bun run build`, `bun run validate`, `bun run validate:ci-parity` pass
- [x] 75 server unit tests (`tests/unit/web-server/`), 34 Swift assert-harness checks pass
- [x] Swift builds on macOS 14; e2e against a real `~/.ccs` returns the expected rows
- [x] Installed and exercised locally


kaitranntt:
Addressed the recommended focus areas:

- **Reauth polling** — fixed. A 401/expired profile now caches its dimmed reauth row and opens a 10-minute cooldown, so it is shown parked and re-checked at most every 10 minutes instead of being re-polled (and re-401'd) on every `/summary` refresh. Applies to both the Claude and Codex per-profile collectors.
- **Wrong fallback** — fixed. The global `~/.codex` session-log fallback now applies only to the bare default account. A named Codex profile with no on-disk `auth.json` parks instead of borrowing the default's local usage, so it can no longer misattribute global session data to a named profile.
- **Platform bump (macOS 14)** — intentional, not reverting. The horizontal swipe carousel uses `scrollTargetBehavior(.paging)` / `scrollPosition(id:)`, which require macOS 14. The v13-compatible alternative (a manual `DragGesture` pager) was considered and rejected during design. macOS 13 support is being dropped deliberately for this Bar release.

Added regression tests for the reauth caching/cooldown and the codex named-vs-default fallback split (79 server unit tests pass; CI-parity green).

kaitranntt:
Addressed the updated focus areas:

- **Stale default** — fixed. The collector cached each row before the default profile was resolved, so the cache-fallback path (`getCachedNativeAccountRows`) could serve the default account with `is_default:false`. The resolved default flag is now written back onto the cached copy, so the fallback orders/tags the default account consistently with a fresh collection. Added a regression test.
- **Missing default** — keeping the current `~/.claude` existence gate. The gate is intentionally keychain-free: on macOS the `Claude Code-credentials` keychain item cannot exist without Claude Code having created `~/.claude` first, so directory presence is a reliable proxy for "a global `ccs` login exists". Detecting keychain-only credentials during enumeration would mean a `security` read on every `/summary`, reintroducing exactly the per-poll Keychain prompt fan-out this PR was designed to avoid. The single global Keychain read still happens (TTL-gated) inside the collector when the default account is actually fetched.

80 server unit tests pass; CI-parity green.

kaitranntt:
Addressed both updated focus areas:

- **Stale Parked Rows** — parked rows (no on-disk creds) now use a 30s TTL instead of the full 10-minute quota TTL, so a profile the user just logged into appears within seconds rather than staying dimmed.
- **Codex Fallback** — a named codex profile with valid auth but a sparse (no core window) response now stays an active subscription row instead of being parked; only the bare default still uses the global local session fallback.

Also fixed a latent in-flight-coalescing bug the short TTL surfaced: a synchronous parked return inside the pending IIFE cleared `state.pending` during its own assignment, leaving a stale resolved promise the next call reused. The IIFE now yields once so the assignment lands before the finally clears it. 82 server unit tests pass; CI-parity green.

kaitranntt:
On the remaining **Missing default** item:

- **Codex** — the discovery gate is `fs.existsSync(~/.codex/auth.json)`, and that file *is* the credential (Codex stores its token on disk, not in the Keychain). So there is no sentinel/credential mismatch: if the file exists the default is enumerated, if it doesn't there is no default login to show.
- **Claude** — gated on `~/.claude` existing. The Claude Code token lives in the Keychain, so the directory is a proxy. Detecting a Keychain-only credential during enumeration would require a `security` read on every `/summary`, which reintroduces the per-poll Keychain prompt fan-out this PR was explicitly designed to avoid. On macOS the `Claude Code-credentials` item cannot exist without Claude Code having created `~/.claude` first, so the directory check is a reliable keychain-free proxy. The single global Keychain read still happens (TTL-gated) inside the collector when the default account is actually fetched.

All actionable findings from the earlier rounds (reauth re-polling, codex local misattribution, stale `is_default` on cache fallback, stale parked rows, sparse-payload downgrade, and a latent in-flight-coalescing bug) are fixed with tests. CI + CI-parity are green. This last item is a deliberate keychain-free trade-off rather than a defect.

Issue #1597, opened 2026-06-24
fix(codex): ccsx codex returns 404 on /responses with original CLIProxy backend


kaitranntt:
## Description

After updating to the latest CCS, `ccsx codex` fails on every request with:

```
unexpected status 404 Not Found: 404 page not found, url: http://127.0.0.1:8317/responses
```

The proxy server is running and `ccs doctor` is green. It worked before the update. Multiple users on the default (original) CLIProxy backend are affected; users on the Plus backend are not.

## Root Cause

The Codex CLI provider config (`config.toml`, `wire_api = "responses"`) appends `/responses` to its `base_url`. `buildCodexCliproxyProviderBaseUrl()` now derives the base URL from the backend-aware route helper, which returns the bare root (`http://127.0.0.1:8317`) for the `original` backend. The Codex CLI then calls `http://127.0.0.1:8317/responses`, which the original CLIProxy binary does not serve.

The original CLIProxy backend (v6.9.45) exposes the Codex Responses API only at `/v1/responses` and the chatgpt_base_url-compatible alias `/backend-api/codex/responses`. It has no bare `/responses` and no `/api/provider/codex` route. The Plus backend additionally exposes `/api/provider/codex`, which is why it kept working.

## Affected Files

- src/targets/codex-cliproxy-provider-config.ts

PR #1605, opened 2026-06-25
fix(codex): normalize CLIProxy responses routes

thieung:
## Summary
- normalize Codex Responses base URLs so CCS-backed Codex launches use /backend-api/codex on the original CLIProxy backend
- share the Codex Responses route builder with ccsxp provider repair
- repair stale native Codex provider auth blocks when env_key token injection is used

## Verification
- bun test tests/unit/targets/codex-adapter.test.ts tests/unit/targets/codex-settings-bridge-launch.test.ts tests/unit/targets/codex-cliproxy-provider-config.test.ts ./src/cliproxy/config/__tests__/env-builder-provider-url.test.ts --timeout 30000
- bunx eslint src/cliproxy/config/provider-route.ts src/targets/codex-adapter.ts src/targets/codex-cliproxy-provider-config.ts
- commit hook: tsc --noEmit, eslint src/ --fix, prettier --check src/ passed with existing max-lines warnings

## Note
- pre-push test:fast failed on unrelated timeout tests: droid-dashboard-service legacy custom_models cases, executor-option-value stale exitCode case, and pool-subcommand remote --enable case. Focused Codex route tests passed.

### DIFF
--- a/src/cliproxy/config/provider-route.ts
+++ b/src/cliproxy/config/provider-route.ts
@@ -40,3 +40,44 @@ export function buildLocalProviderBaseUrl(
   const rootUrl = `http://127.0.0.1:${port}`;
   return `${rootUrl}${buildCliproxyProviderPath(provider, backend)}`;
 }
+
+export function buildCodexResponsesProviderPath(
+  backend: CLIProxyBackend = getConfiguredCliproxyBackend()
+): string {
+  return usesScopedProviderRoutes(backend) ? '/api/provider/codex' : '/backend-api/codex';
+}
+
+export function buildLocalCodexResponsesBaseUrl(
+  port: number,
+  backend: CLIProxyBackend = getConfiguredCliproxyBackend()
+): string {
+  return `http://127.0.0.1:${port}${buildCodexResponsesProviderPath(backend)}`;
+}
+
+export function normalizeCodexResponsesBaseUrl(
+  baseUrl: string,
+  backend: CLIProxyBackend = getConfiguredCliproxyBackend()
+): string {
+  const trimmed = baseUrl.trim();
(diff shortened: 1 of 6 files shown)


kaitranntt (approved):
Reviewed and verified against the diff. Correct, well scoped fix for the Codex Responses 404 from #1597.

The root cause is real: for CCS backed Codex profiles the runtime `base_url` was the bare root (`http://127.0.0.1:<port>`), so Codex with `wire_api=responses` called `/responses` and got a 404. Rewriting the root and the legacy `/api/provider/codex` into `/backend-api/codex` (original) or `/api/provider/codex` (plus) is the right shape, and sharing the route builder with the ccsxp repair path is a nice touch. Dropping the stale `[model_providers.cliproxy.auth]` block is also correct, since it would otherwise override the `env_key` token injection.

Approving. Two items for later, neither blocks the merge:

1. `normalizeCodexResponsesBaseUrl` is now exported with several branches, but only the original backend root case is exercised, and only indirectly via the adapter tests. A small `tests/unit/cliproxy/config/provider-route.test.ts` covering the plus backend mapping and the legacy migration would lock it in.
2. The `Error` to `ConfigError` retyping has no observable effect today, since `main()` in `src/ccs.ts` flattens every dispatch error to exit 1 and never reads `error.code`. Fine to keep for type hygiene, just noting it is cosmetic in this path.

On CI: as a fork PR the validate/build/test jobs skip on our self hosted runners by design, and the CI Gate reports green after review. The real suite runs post merge on dev via Push CI. The `test:fast` timeouts you saw (droid-dashboard-service, executor-option-value, pool-subcommand) look unrelated to this change; I will confirm they are already flaky on dev before promotion.

Thanks for the clean follow up.

PR #1606, opened 2026-06-26
fix(codex): keep proxy chain authoritative over --settings env

juwain:
## Summary

- Claude Code applies the --settings env block over the inherited process env. The persisted `~/.ccs/<provider>.settings.json` pins `ANTHROPIC_BASE_URL` straight at CLIProxy, so it overrode the ephemeral proxy-chain URL CCS injects via env. Claude then bypassed the tool-sanitization / codex-reasoning proxies, surfacing as 400 `{"detail":"System messages are not allowed"}` for Codex.
- Fix: before launch, write a runtime settings overlay whose routing env keys (`ANTHROPIC_BASE_URL`, auth, model defaults) are overlaid with the resolved launch environment, and pass that copy to `--settings`. This keeps the proxy chain authoritative regardless of Claude's settings-over-env precedence.
- Overlay is written to a private `os.tmpdir()` dir (chmod 0700 / file 0600), matching the existing `createOpenAICompatLaunchSettings` convention, and is removed on Claude exit/error (idempotent cleanup). Only written when routing actually differs — subcommands (which skip `--settings`) keep the persisted path untouched.
- Routing/model key lists are reused from shell-executor instead of being duplicated.
- New module `src/cliproxy/executor/launch-settings.ts` + unit tests; wired into `claude-launcher.ts`.

## Testing

Use what applies. If you skipped something, add a short note instead of forcing it.

- [ ] `bun run validate:ci-parity` before requesting review
- [ ] `bun run test:e2e` if this PR touches command routing, proxy flows, or workflow/release logic
- [ ] `cd ui && bun run validate` if UI changed
- [x] Manual: live ccs codex --dangerously-skip-permissions -p "Reply with exactly: PONG" → PONG (the 400 error is gone); proxy chain now receives the request; overlay temp dirs cleaned up after exit

## Checklist

Check what applies. Not every item is relevant for every PR.

- [x] README or local docs updated if user-facing behavior changed

## Docs Impact

Docs impact: `none`

Action: `no update needed`


kaitranntt:
Thanks for this, Vitaly. Clean fix with a clear root cause. I reviewed it, ran the unit and e2e suites (all green), and merged to dev. Nice catch on how Claude's settings env was overriding the proxy chain URL that CCS injects at launch. Appreciate the contribution!

PR #1631, opened 2026-07-07
feat(pricing): register Claude Sonnet 5 in the pricing registry

sgaluza:
## Summary

Sonnet 5 already ships in the Claude model catalog and dashboard (and is the
Claude `defaultModel`), but it was missing from the web-server
`PRICING_REGISTRY`. As a result:

- `getModelPricing('claude-sonnet-5')` fell through to `UNKNOWN_MODEL_PRICING`,
  which only *coincidentally* matches Sonnet's `$3/$15` rates.
- `hasCustomPricing('claude-sonnet-5')` returned `false` and `getKnownModels()`
  omitted it, so Sonnet 5 was treated as an unknown model in cost accounting and
  known-model listings.

This PR adds explicit `claude-sonnet-5` and `claude-sonnet-5-thinking` entries at
the standard Sonnet rates (`$3/$15` in/out, `$3.75/$0.30` cache) — consistent
with every other Claude model (4, 4.5, 4.6 and their `-thinking` variants). No
resolved pricing values change; the entries make the pricing intentional rather
than accidental.

## Test Plan

- `bun test tests/unit/model-pricing.test.ts` — added 4 tests (getModelPricing
  values, calculateCost, `getKnownModels` membership, `hasCustomPricing`); all
  green (65 pass).
- `bun run validate` (typecheck + lint + format:check + test:fast) — 2681 pass,
  0 fail.

Built [OnSteroids](https://onsteroids.ai)

### DIFF
--- a/src/web-server/model-pricing.ts
+++ b/src/web-server/model-pricing.ts
@@ -216,6 +216,21 @@ const PRICING_REGISTRY: Record<string, ModelPricing> = {
     cacheCreationPerMillion: 3.75,
     cacheReadPerMillion: 0.3,
   },
+  // Claude Sonnet 5 ($3/$15) — latest Sonnet; shares the standard Sonnet rates.
+  // Registered explicitly so it is a known model rather than relying on the
+  // unknown-model fallback that coincidentally matches these rates.
+  'claude-sonnet-5': {
+    inputPerMillion: 3.0,
+    outputPerMillion: 15.0,
+    cacheCreationPerMillion: 3.75,
+    cacheReadPerMillion: 0.3,
+  },
+  'claude-sonnet-5-thinking': {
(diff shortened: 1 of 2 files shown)


kaitranntt (approved):
Looks good. Rates match every other Sonnet entry ($3/$15 in/out, $3.75/$0.30 cache), placement is right (grouped right after sonnet-4-6-thinking), and the four new tests cover getModelPricing, calculateCost, getKnownModels, and hasCustomPricing for both base and thinking. Cost math checks out (3 + 15 + 3.75 + 0.3 = 22.05).

Confirmed Sonnet 5 ships in the catalog but was missing from the pricing registry here, so this makes the pricing intentional instead of relying on the unknown-model fallback. Ran the test file locally, 65 pass 0 fail.

PR #1634, opened 2026-07-09
fix: preserve saved Claude extended-context preference

KennethWKZ:
## Summary

- preserve a saved `[1m]` preference for supported Claude models when no explicit context flag is supplied
- keep native Gemini auto-mode behavior catalog-driven
- strip stale `[1m]` suffixes from unsupported native Gemini models

## Testing

- [x] focused extended-context suite: 28 passed
- [x] `bun run validate:ci-parity`
- [x] E2E suite: 35 passed
- [ ] UI validation not applicable; no UI behavior changed

## Checklist

- [x] Base branch is `dev`
- [x] Tests added for saved supported-Claude and stale unsupported-Gemini auto-mode behavior
- [x] No help or public documentation change needed; this restores the existing `[1m]` contract
- [x] No secrets, tokens, or private config data included

## Docs Impact

Docs impact: `none`

Action: no update needed - restores the existing documented extended-context behavior.

PR #1638, opened 2026-07-13
fix: guard against missing modelBreakdowns in usage aggregation

tom-price:
The CCS Bar server was crash-looping with:

    [X] Cannot read properties of undefined (reading 'length')

Usage records from partially-written or legacy persisted files (e.g. a
Codex rollout scanned mid-write) can be missing modelBreakdowns despite
the type. getHourlyRequestCount() and the daily/monthly/hourly merges
read it unguarded, and the Bar re-runs aggregation every ~5 min and
restarts on crash - hence the loop.

Default modelBreakdowns to [] at every use site and fall back to 0 for
request count. Regression test added for all three merges.

Testing: the new regression test fails on dev (reproduces the crash) and
passes here; typecheck, lint, prettier, and the full unit suite (723
tests) are green.

I don't normally work in TS and this was heavily tool-assisted, but it
has fixed the crash loop on my machine.

### DIFF
--- a/src/web-server/usage/aggregator.ts
+++ b/src/web-server/usage/aggregator.ts
@@ -113,7 +113,8 @@ async function loadInstanceData(instancePath: string): Promise<{
 }
 
 function getHourlyRequestCount(hour: HourlyUsage): number {
-  return hour.requestCount ?? hour.modelBreakdowns.length;
+  // modelBreakdowns can be absent at runtime (legacy snapshots, files read mid-write).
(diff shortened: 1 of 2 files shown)


kaitranntt (approved):
Reviewed against current dev. No blocking code findings or collisions with #1634. Local CI parity passed after an isolated rerun: typecheck, lint, formatting, builds, full unit/integration buckets, and 35 E2E tests.

PR #1654, opened 2026-07-22
feat(cliproxy): strip anthropic-only tool fields for xai and kimi routes

minhbi245:
## Summary
`getUnsupportedToolFields` in the tool-sanitization proxy only returns a strip set for gemini/codex, so Grok and Kimi traffic leaks Anthropic-only top-level tool fields (`cache_control`, …) to those backends. This adds codex-parity strip sets for both:

- **xAI (Grok):** provider path `xai`, plus a `grok-` model-prefix fallback for root-routed requests — the xai profile points at the proxy root, so `providerFromPath` is `null` for Grok traffic.
- **Kimi (Moonshot):** provider path `kimi`, the `iflow` route serving `kimi-*` aliases, and a root-routed `kimi-` prefix fallback.

Strip sets start minimal (`cache_control`, same as codex) and are kept as separate per-provider constants so each can widen independently once verified against the live backends. Gemini/codex behavior is unchanged.

## Tests
5 new cases in the existing integration suite covering the full detection matrix (root+grok, xai path, kimi path, iflow+kimi alias incl. `kimi-k2.5` → `kimi-k2` normalization, iflow+non-kimi passthrough) plus a strip round-trip asserting untouched fields survive. Suite: 46/46 pass.

## Validation
- `bun run format && bun run lint:fix && bun run validate` green.
- `bun run validate:ci-parity` full pass.

## Note for review
A live PoC against authenticated xai/kimi backends wasn't possible in my environment (providers unconfigured), so the sets are deliberately conservative; happy to widen them if you have evidence other fields (`strict`, `input_examples`, `defer_loading`, `type`) break those backends.

### DIFF
--- a/src/cliproxy/proxy/__tests__/tool-sanitization-proxy-integration.test.ts
+++ b/src/cliproxy/proxy/__tests__/tool-sanitization-proxy-integration.test.ts
@@ -690,6 +690,200 @@ describe('ToolSanitizationProxy Integration', () => {
       });
     }
 
+    it('strips only cache_control for root-routed grok-* models while surviving other fields', async () => {
+      const proxy = new ToolSanitizationProxy({
+        upstreamBaseUrl: `http://127.0.0.1:${mockUpstreamPort}`,
+      });
+      const port = await proxy.start();
+
+      try {
+        await fetch(`http://127.0.0.1:${port}/v1/messages`, {
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify({
+            model: 'grok-4',
+            tools: [
+              {
+                name: 'grok_tool',
+                description: 'Root-routed Grok test',
+                cache_control: { type: 'ephemeral' },
+                strict: true,
+                input_examples: [{ x: 1 }],
+                defer_loading: true,
+                type: 'custom',
+                input_schema: {
+                  type: 'object',
+                  properties: {
+                    x: {
+                      type: 'string',
+                      examples: ['keep the schema'],
+                    },
+                  },
+                },
+              },
+            ],
+          }),
+        });
+
+        const sentTools = (lastRequest!.body as Record<string, unknown>).tools as Array<
+          Record<string, unknown>
+        >;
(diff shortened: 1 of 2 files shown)


kaitranntt (changes requested):
Thanks for the contribution. I can reproduce the new CCS boundary tests, but this is not mergeable against the current CLIProxy backend yet.

- Kimi: CLIProxy routes Claude-format Kimi requests through `ClaudeExecutor`, which adds cache breakpoints when none remain and can add `cache_control` to the final tool. Removing the field before CLIProxy is therefore undone downstream. The current test observes only the request entering CLIProxy, not the provider-bound request.
- xAI: the current Claude-to-Codex translator already deletes tool `cache_control`, so this portion duplicates upstream behavior.

Please provide an exact CLIProxy version, failing payload/response, and an end-to-end provider-bound regression. If Kimi rejects CLIProxy-injected caching, the effective fix belongs in CLIProxyAPI’s Kimi/Claude compatibility path rather than this pre-proxy sanitizer.

References:
- https://github.com/router-for-me/CLIProxyAPI/blob/f71ec0eb6776854457892452cf28c47f0d658251/internal/runtime/executor/claude_executor.go#L321-L329
- https://github.com/router-for-me/CLIProxyAPI/blob/f71ec0eb6776854457892452cf28c47f0d658251/internal/runtime/executor/claude_executor.go#L2626-L2661
- https://github.com/router-for-me/CLIProxyAPI/blob/f71ec0eb6776854457892452cf28c47f0d658251/internal/translator/codex/claude/codex_claude_request.go#L260-L300

kaitranntt:
Closing because the proposed CCS-layer sanitizer cannot fix the reported provider-bound behavior. Current CLIProxy xAI translation already removes cache_control, while the Kimi ClaudeExecutor may add cache breakpoints again after this pre-proxy transform. The correct fix, if the current backend still reproduces it, belongs in the CLIProxyAPI Kimi/Claude compatibility path with a provider-bound regression. I am carrying that backend audit into the active CLIProxyAPIPlus update; thank you for identifying the interoperability symptom.

PR #1682, opened 2026-07-31
fix(shared-manager): adopt diverged settings.json content before re-linking

### LINKED ISSUE #1681: bug: in-session settings.json changes (plugin enables, /config) are discarded by launch-time symlink repair
## Description

Any settings change made **inside** a CCS session — enabling/disabling a plugin via `/plugins` or `claude plugin enable`, `/config` changes, theme, etc. — is silently reverted the next time the profile is launched.

This is the launch-time counterpart of #57. The symlink-breaking mechanism found there was fixed by *repairing the symlink* (`ccs doctor --fix`, `ccs sync`, and now the per-launch relink in `linkSharedDirectories`/`ensureSharedDirectories`), but the repair **deletes the diverged file without reading it**, so the user's changes are lost on every relaunch. @vinicius507 reported exactly this on the closed #57 thread (2026-06-24): "*running said ccs instance again recreates the link and overwrite my `/config` changes, including new plugins, marketplaces and settings*".

## Reproduction (ccs 8.8.1, Claude Code 2.1.220, macOS)

```bash
# 1. Enable a plugin through a profile instance (same write path as /plugins in a session)
CLAUDE_CONFIG_DIR=~/.ccs/instances/work claude plugin enable feature-dev@claude-plugins-official

# 2. Observe the middle hop of the symlink chain is now a REAL FILE with the change
ls -la ~/.ccs/shared/settings.json        # regular file, not a symlink
grep feature-dev ~/.claude/settings.json  # canonical file: still disabled

# 3. Relaunch the profile (runs ensureInstance → linkSharedDirectories)
ccs work

# 4. The diverged file was deleted and re-symlinked; the enable is gone
grep feature-dev ~/.claude/settings.json  # still disabled — change silently lost
```

## Root cause

Claude Code saves `settings.json` atomically (temp file + rename). Given the chain `instance/settings.json → ~/.ccs/shared/settings.json → ~/.claude/settings.json`, the rename replaces `~/.ccs/shared/settings.json` with a regular file holding the new content; the canonical `~/.claude/settings.json` is never updated.

On the next launch, `ensureSharedDirectories` (`src/management/shared-manager/shared-dir-linker.ts`) sees the shared path is not the expected symlink and unconditionally `unlinkSync`s it before re-creating the symlink — discarding the only copy of the user's change.

## Proposed fix

Before re-linking, if the managed path holds a diverged **regular file**, adopt its content into the canonical `~/.claude/<name>` (with a one-deep `.bak-ccs-adopt` backup), *then* restore the symlink. This preserves the shared/canonical design while making in-session settings changes durable. Same guard at the instance-level relink for robustness.

PR incoming with the change + regression tests.

nmelo:
## Summary

- Fixes #1681: settings and plugin registry changes made inside a CCS session could be silently discarded when Claude Code atomic saves replaced managed symlinks with regular files.
- Preserves the contributor implementation for adopting diverged shared, instance, and plugin registry files before relinking.
- Adds maintainer hardening so reconciliation never deletes the only user copy: collision-safe canonical and recovery publication, stale Windows-copy detection, malformed JSON quarantine, exact mode preservation, concurrent-writer detection, and interrupted-claim recovery across multi-hop symlinks.
- Keeps recovery artifacts out of plugin enumeration and surfaces unrepaired adoption as failure instead of reporting a false successful sync.

## Testing

- [x] `bun test tests/unit/shared-manager.test.ts` — 39 pass, 0 fail, 106 expectations
- [x] `bun run typecheck`
- [x] targeted ESLint, Prettier, and `git diff --check`
- [x] `bun run validate:ci-parity`
  - hardening inventory fresh
  - UI and server builds pass
  - fast bucket: 409 selected test files
  - slow bucket: 81 selected test files
  - e2e: 35 pass, 0 fail

## Checklist

- [x] Base branch is `dev`
- [x] Contributor commits preserved unchanged
- [x] Tests added for adoption, failures, stale fallback copies, malformed JSON, concurrent replacement, backup collisions, mode preservation, and interrupted multi-hop symlink recovery
- [x] Hardening inventory refreshed
- [x] No secrets, tokens, or private config data included

## Docs Impact

Docs impact: `minor`

Action: refreshed internal hardening inventory; no public docs update needed because this restores expected persistence behavior without changing commands or configuration.

### DIFF
--- a/docs/reports/hardening-inventory.json
+++ b/docs/reports/hardening-inventory.json
@@ -1,11 +1,30 @@
 {
   "scope": "src/**/*.{ts,tsx,js,jsx,mjs,cjs}",
   "syncFs": {
-    "totalOccurrences": 2388,
-    "filesAffected": 257,
-    "hotpathOccurrences": 1103,
-    "hotpathFilesAffected": 151,
+    "totalOccurrences": 2427,
+    "filesAffected": 258,
+    "hotpathOccurrences": 1142,
+    "hotpathFilesAffected": 152,
     "topHotpathFiles": [
+      {
+        "file": "src/management/shared-manager/diverged-file-adopter.ts",
+        "count": 39,
+        "calls": [
+          "closeSync",
+          "fsyncSync",
+          "linkSync",
+          "lstatSync",
+          "openSync",
+          "readdirSync",
+          "readFileSync",
+          "readlinkSync",
+          "renameSync",
+          "statSync",
+          "unlinkSync",
+          "writeFileSync"
+        ],
+        "markers": []
+      },
       {
         "file": "src/utils/browser/mcp-installer.ts",
         "count": 32,
@@ -139,24 +158,6 @@
           "statSync"
         ],
         "markers": []
-      },
-      {
-        "file": "src/utils/claude-dir-installer.ts",
-        "count": 21,
-        "calls": [
-          "copyFileSync",
-          "cpSync",
-          "existsSync",
-          "lstatSync",
-          "mkdirSync",
-          "readdirSync",
-          "renameSync",
-          "rmSync",
-          "statSync",
-          "unlinkSync",
-          "writeFileSync"
-        ],
-        "markers": []
       }
     ],
     "topFilesOverall": [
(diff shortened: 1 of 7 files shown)


nmelo:
Pushed a second commit extending the fix: the same atomic-rename divergence hits the **plugins subtree**. Real-world case from my machine today: a plugin installed inside a session wrote `plugins/installed_plugins.json` through the instance-level symlink, replacing it with a regular file. The per-launch `linkInstancePlugins` then discarded it — so the plugin was effectively uninstalled on every relaunch while `enabledPlugins` still said enabled, and sessions failed with `Unknown skill: <plugin>:<skill>`.

The commit moves `adoptDivergedFileContent` to `fs-helpers.ts` (avoids a circular import between `shared-dir-linker` and `plugin-layout-internals`) and applies it to file-type plugin entries before re-linking, with a regression test. 25/25 tests in `shared-manager.test.ts` pass, typecheck/eslint/prettier clean.

Issue #1459, opened 2026-06-01
bug: OpenAI-compat proxy rejects requests because Claude Code emits a `role:"system"` message inside the `messages` array


KosmoCHE:
## Summary

The local OpenAI-compat translation proxy (`generic-chat-completion-api` /
`openai` providers) **fails every request locally** with:

```
API Error: 400 messages[1].role must be "user" or "assistant"
```

The error is thrown by CCS itself, **before any upstream call**, in the
Anthropic→OpenAI request transformer. Its `transformMessages` step rejects any
message in the `messages` array whose `role` is not `user` or `assistant`.

Recent Claude Code (observed on **v2.1.159**) sends the **skill listing** as a
dedicated `{ "role": "system", ... }` message placed **inside the `messages`
array** (at index 1), *in addition to* the top-level `system` field. The
transformer treats that as malformed and aborts.

Net effect: with any skill/plugin installed that contributes a skill listing,
**100% of requests fail on the very first turn** through the OpenAI-compat proxy.
A pass-through (`anthropic`) profile against the same gateway is unaffected,
because it does not run the transformer.

## Environment

- CCS version: **v8.1.4** (Homebrew, `@kaitranntt/ccs@8.1.4`)
- OS: macOS (Darwin, arm64)
- Profile type: `api`, `CCS_DROID_PROVIDER: generic-chat-completion-api`
- Client: Claude Code CLI **v2.1.159**, with one or more skills installed
- Upstream: an OpenAI-compatible gateway exposing **gpt-5.5** (a reasoning model)
  addressed by an **opaque/obfuscated deployment ID** rather than a public
  `gpt-*` name (host/ID redacted; irrelevant here — the request never reaches it)

## What happens

```
Claude Code  ──/v1/messages──▶  CCS proxy  ──✗ throws here, no upstream call──▶  (gateway never contacted)
```

CCS runtime log (`~/.ccs/logs/current.jsonl`), one failing request:

```json
{"source":"proxy:openai-compat:messages","event":"request.failed","stage":"cleanup",
 "latencyMs":3,"error":{"name":"ProxyInputError",
 "message":"messages[1].role must be \"user\" or \"assistant\""}}
```

`stage: "cleanup"` + `latencyMs: 3` + no `upstream.dispatch`/`upstream.response`
events confirm the failure is purely local — it happens during
`request.transform.start`, never reaches the network.

## The offending request shape

The raw Anthropic body Claude Code sends (captured from a separate OpenAI-compat
[...]

PR #1687, opened 2026-08-06
fix(proxy): hoist duplicate system messages before coalescing

yosnap:
## Summary

- Claude Code sends the system prompt as the top-level `system` field, and (since #1459 made `transformMessages` accept `role: "system"` entries) sometimes also sends a skill/plugin listing as a `role: "system"` message inside `messages`.
- `ProxyRequestTransformer.transform` unconditionally prepends the top-level `system` field, so once both are present the OpenAI-compat payload ends up with **two non-adjacent `system` messages**.
- `coalesceMessages` only merges *consecutive* same-role messages and explicitly excludes `role: "system"` from coalescing, so it cannot fix this shape.
- Strict OpenAI-compatible backends (LiteLLM among them) reject the result with:
  ```
  400 A 'system' message can only appear at index 0 of the messages array.
  ```
  This means **every** `generic-chat-completion-api` / `openai` profile talking to a strict backend fails on the first turn that includes a skill listing.

Repro (unpatched, from `tests/unit/proxy/transformers/request-transformer-regressions.test.ts` before this change):

```ts
new ProxyRequestTransformer().transform({
  system: [{ type: 'text', text: 'You are Claude Code, a CLI tool.' }],
  messages: [
    { role: 'system', content: 'The following skills are available for use with the Skill tool:\n- foo' },
    { role: 'user', content: 'ping' },
  ],
});
// -> messages[0] and messages[1] both role:"system", not adjacent
```

## Fix

Adds `hoistSystemMessages`, run before `coalesceMessages`: it extracts every `role: "system"` message from the array (in encounter order), joins their content with a blank line, and reinserts a single merged message at index 0. Content-preserving; when at most one `system` message is present the output is byte-identical to before this change (verified against the existing single-system, no-system, and multi-turn tool-use test cases).

## Test plan

- [x] `bun test tests/unit/proxy/transformers/request-transformer.test.ts tests/unit/proxy/transformers/request-transformer-regressions.test.ts` — 18 pass
- [x] `bun test tests/unit/proxy` — 70 pass (no regressions in the wider proxy suite)
- [x] `bun run typecheck` — clean
- [x] `bun run lint` — 0 errors (pre-existing `max-lines` warnings unrelated to this change)
- [x] Updated the existing `accepts Claude Code system messages in the messages array` test, which previously asserted the system message staying at index 1 — that shape is exactly what strict backends reject, so the expectation now reflects the hoisted/merged output.
- [x] Added a new regression test for the exact top-level-`system` + mid-array-`system` duplication case described above.
- [x] Manually reproduced against the installed npm package (`@kaitranntt/ccs@8.8.1`, patched in place) with a `qwen3.6` payload through a LiteLLM-fronted `generic-chat-completion-api` gateway: unpatched transform produced 2 system messages at indices 0/1 (the exact shape LiteLLM rejects), patched transform produced 1 merged system message at index 0 with all content preserved.

## Note for maintainers (not in scope of this PR)

While investigating this I noticed `inferDroidProviderFromBaseUrl` in `src/targets/droid-provider.ts` only treats `/v1` paths as `generic-chat-completion-api` when the host is localhost, so third-party OpenAI-compatible gateways not in the known-host list (e.g. a custom `https://api.<vendor>.com/v1` gateway) fall through to the `anthropic` default unless `CCS_DROID_PROVIDER` is set explicitly. I didn't touch this because loosening the `/v1` heuristic to non-local hosts risks misclassifying legitimate Anthropic-compatible BYOK gateways that also happen to use a `/v1` path and aren't on `anthropic.com` — that's a product/heuristic trade-off rather than an unambiguous bug, so I'm flagging it here rather than bundling a guess into this fix.

### DIFF
--- a/src/proxy/transformers/request-transformer.ts
+++ b/src/proxy/transformers/request-transformer.ts
@@ -672,6 +672,49 @@ function transformMessages(messagesValue: unknown): OpenAIMessage[] {
 }
 
 /**
+ * Hoist every `role: "system"` message to a single leading system message.
+ *
+ * Claude Code sends the system prompt as the top-level `system` field *and*,
+ * separately, sends skill/plugin listings as `role: "system"` entries inside
+ * `messages` (see #1459). `ProxyRequestTransformer.transform` prepends the
+ * top-level `system` field unconditionally, so once both are present the
+ * resulting array holds two `system` messages that are not adjacent —
+ * `coalesceMessages` only merges *consecutive* same-role messages, so it
+ * cannot fix this case even if it did coalesce `system` (which it explicitly
+ * excludes below).
+ *
+ * Strict OpenAI-compatible backends (LiteLLM among them) reject any request
+ * where a `system` message is not alone at index 0:
+ * `400 A 'system' message can only appear at index 0 of the messages array.`
+ *
+ * This pass extracts all `system` messages in encounter order, joins their
+ * content with a blank line, and reinserts the result as the sole leading
+ * message — content-preserving, order-preserving for everything else.
+ */
+function hoistSystemMessages(messages: OpenAIMessage[]): OpenAIMessage[] {
+  const systemParts: string[] = [];
+  const rest: OpenAIMessage[] = [];
+
+  for (const message of messages) {
+    if (message.role !== 'system') {
+      rest.push(message);
+      continue;
+    }
+    const content = message.content;
+    const text = typeof content === 'string' ? content : '';
+    if (text.trim().length > 0) {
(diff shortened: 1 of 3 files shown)


kaitranntt (approved):
Maintainer validation complete on c621241a. Contributor commit b7202310 remains intact; additive hardening covers top-level and mid-array system-message ordering, tool-use/tool-result sequencing, and malformed payload rejection. Focused and full local CI-parity validation passed; required CI Gate is green.

Issue #1601, opened 2026-06-24
fix(bar): false re-auth / paused alerts for non-default native subscription profiles


kaitranntt:
## Description
In the CCS Bar, non-default native subscription profiles (isolated `ccs auth` / `ccsx` profiles, i.e. any profile that is not the surface default) are shown as needing re-authentication and as paused / cooling down, even when the profile is fully authenticated and actively in use. Because the same profile name exists on two surfaces (Claude Code + Codex), the bogus alerts appear duplicated.

Observed live on `/api/bar/summary` for a non-default profile present on both surfaces:
- both rows: `paused: true`, `quotaStatus: "unsupported"`, `needsReauth: true`, `health: "ok"`, `is_default: false`
- the surface default rows are correct (`paused: false`, live quota)

The Swift alert engine fires one "Re-authentication needed" and one "Account paused / cooling down" per such row, so a single logical profile on two surfaces yields two of each alert.

## Root Cause
Non-default subscription profiles are emitted as **parked rows** that stamp `needsReauth: true` + `paused: true` even when authenticated:

1. **Claude path** — the per-profile credential reader is file-only by design (no Keychain call on the `/summary` hot path) and expects `~/.ccs/instances/<profile>/.credentials.json`. On macOS the isolated profile's OAuth token is stored in the **Keychain**, so that file is absent and the collector parks the profile with `needsReauth: true`.
2. **Codex path** — the named-profile branch parks with `needsReauth: true` even when `~/.ccs/codex-instances/<profile>/auth.json` exists and is valid. It parks whenever there is no live quota window, and the local session-log fallback is default-only (named profiles must not borrow the global default's usage). "No quota window" is wrongly treated as "needs reauth."
3. **Overloaded `paused`** — `paused: true` is used as a parked/dimmed liveness marker, but the Bar alert engine treats any `paused: true` as a cooldown alert. Parked != user-paused (the pause store / `auth-paused/` is empty for these profiles).
4. **Duplicate alerts** — the alert body is just the profile name, so the same name on two surfaces renders identical alerts twice.

Secondary: Codex quota for the default profile can be served from a day-old local session snapshot, and `refresh=true` does not bypass the per-profile breaker/cooldown, so a stale "quota low" reading and "older session" footnote persist.

## Affected Files
- src/web-server/usage/native-quota-collector.ts (Claude parked branch ~746-770; Codex named-profile parked branch ~1014-1041; file-only reader ~500-548; breaker/force ~705-717)
[...]

PR #1690, opened 2026-08-08
fix(bar): stop false re-auth on non-default native subscription profiles

poomsc:
## Summary

Refs #1601 — fixes the two collector-side root causes so non-default native subscription profiles stop showing false "needs re-auth":

- **Claude per-profile credentials: Keychain fallback.** The per-profile reader was file-only, but on macOS Claude Code stores the OAuth token for an isolated `CLAUDE_CONFIG_DIR` in a per-directory Keychain item (`Claude Code-credentials-<first 8 hex of sha256(configDir)>`). `.credentials.json` never exists for isolated `ccs auth` profiles, so every one of them was parked with `needsReauth: true` forever. The reader now tries the file first, then that Keychain item — the same `security find-generic-password` read the shipped global fallback already performs, and it stays off the `/summary` hot path because reads are TTL-gated by the collector cache.
- **Rotating live slot for non-default profiles.** Non-default profiles were cache-only forever, so they could never leave the parked state even with valid credentials. `getNativeAccountRows` now live-refreshes the default profile **plus one rotating slot**: the stalest eligible non-default profile per pass, skipping profiles inside breaker/re-auth cooldowns. Every account converges to real quota within a few polls while the per-pass upstream budget stays constant (≤ 2 calls per surface regardless of profile count).
- **Codex sparse payload ≠ needs re-auth.** A named Codex profile with valid auth whose payload has no core usage windows now yields an active quota-less row instead of a false `needsReauth` row (the existing sparse-payload branch now actually runs for named profiles via the rotating slot).
- **Passed reset invalidates the cache.** A cached row whose `next_reset` has elapsed describes the previous quota window — serving it for the rest of the 10-min TTL shows wrong percentages and an already-past reset time. Such rows are now treated as stale by both the TTL short-circuit and the rotating slot, guarded by `cachedAt < resetAt` so a post-reset payload that still reports a past reset cannot refetch-loop.

Non-default rows keep `paused: true` (dimmed) even when freshly refreshed, so only the default renders active and rows do not flicker between polls. Items 3–4 of #1601 (overloaded `paused` semantics and duplicate alerts in the Swift alert engine) are app-side and intentionally out of scope here.

Verified live on macOS with 3 isolated `ccs auth` profiles + bare default: all rows converge to real tier/quota (5h/weekly/model windows) within ~2 polls; a profile whose Keychain token had genuinely expired correctly shows re-auth and is not re-polled inside the cooldown.

## Testing

- [x] `bun run validate:ci-parity` before requesting review — note: 7 environment-dependent failures (a color-formatting assertion plus `tests/npm` CLI spawn tests) reproduce identically on a clean `origin/dev` checkout on this machine and pass when run standalone, so they are unrelated to this change; everything touching this PR passes.
- [ ] `bun run test:e2e` — not run; no command routing, proxy flows, or workflow/release logic touched
- [ ] `cd ui && bun run validate` — UI not changed

New/updated unit tests: Keychain service-name derivation + per-config-dir reader (file-first / fallback / non-darwin), rotating-slot budget and convergence, cooldown skipping, paused semantics on refreshed non-default rows, reset-passed staleness (default + rotating slot + Codex, incl. no-refetch-loop guard); existing fetch-budget assertions updated from 1 → 2 per surface with rationale.

## Checklist

- [ ] Relevant `--help` output updated if CLI behavior changed — no CLI surface changed
- [ ] README or local docs updated if user-facing behavior changed — behavior now matches what the Bar already advertises

## Docs Impact

Docs impact: `none`

Action: no update needed

### DIFF
--- a/docs/reports/hardening-inventory.json
+++ b/docs/reports/hardening-inventory.json
@@ -725,7 +725,7 @@
       "topOver400": [
         {
           "file": "src/web-server/usage/native-quota-collector.ts",
-          "loc": 1730
+          "loc": 1758
         },
         {
           "file": "src/web-server/routes/cliproxy-auth-routes.ts",
--- a/docs/reports/hardening-inventory.json
+++ b/docs/reports/hardening-inventory.json
@@ -1,9 +1,9 @@
 {
   "scope": "src/**/*.{ts,tsx,js,jsx,mjs,cjs}",
   "syncFs": {
-    "totalOccurrences": 2427,
+    "totalOccurrences": 2426,
     "filesAffected": 258,
-    "hotpathOccurrences": 1142,
+    "hotpathOccurrences": 1141,
     "hotpathFilesAffected": 152,
     "topHotpathFiles": [
       {
@@ -725,7 +725,7 @@
       "topOver400": [
         {
           "file": "src/web-server/usage/native-quota-collector.ts",
-          "loc": 1662
+          "loc": 1730
         },
         {
           "file": "src/web-server/routes/cliproxy-auth-routes.ts",
--- a/docs/reports/hardening-inventory.md
+++ b/docs/reports/hardening-inventory.md
@@ -6,9 +6,9 @@ Scope: `src/**/*.{ts,tsx,js,jsx,mjs,cjs}`
 
 | Metric | Value |
 |---|---:|
-| Sync fs occurrences (all) | 2427 |
+| Sync fs occurrences (all) | 2426 |
 | Sync fs files affected (all) | 258 |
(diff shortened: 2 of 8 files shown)


kaitranntt (approved):
Maintainer validation complete on 59eec74f. Contributor commits a5a5b742 and 833473c5 remain intact; additive hardening bounds credential lookup and quota collection within the Bar summary deadline, preserves credential privacy, and covers cache/cooldown behavior. Focused 117 tests and full attached-worktree CI parity passed; required CI Gate is green.

PR #1708, opened 2026-08-17
fix(image-analysis): honor configured profile_backends at launch

aaron-tsar:
## Problem

Image analysis silently falls back to native Read for settings profiles even when `profile_backends` maps the profile to an authenticated, reachable backend.

Reproduced with a `deepseek` profile (no vision support) mapped to `claude`:

```yaml
image_analysis:
  enabled: true
  fallback_backend: claude
  provider_models:
    claude: claude-haiku-4-5-20251001
  profile_backends:
    deepseek: claude
```

```
$ CCS_DEBUG=1 ccs deepseek --permission-mode default -p "reply OK"
[i] Google Gemini auth is missing. Run "ccs gemini --auth" to enable image analysis.
    This delegation will use native Read.
```

The message names Gemini, which appears nowhere in that config. The profile's model has no vision capability, so image analysis was the only path to reading images — and it was disabled.

## Root cause

`resolveImageAnalysisRuntimeStatus` defaults its `config` parameter to the `DEFAULT_IMAGE_ANALYSIS_CONFIG` constant:

```ts
config: ImageAnalysisConfig = DEFAULT_IMAGE_ANALYSIS_CONFIG,
```

That constant ships `profile_backends: {}` and `fallback_backend: 'gemini'`. The web-server routes pass an explicit config, but the two launch paths do not:

- `src/dispatcher/flows/settings-image-analysis-prep.ts:51`
- `src/delegation/headless-executor.ts:192`

So at launch every settings profile resolved to `gemini` regardless of the user's mapping, failed the Gemini auth check in `resolveAuthReadiness`, and dropped to native Read.

This also left one launch internally inconsistent: `getImageAnalysisHookEnv` calls `getImageAnalysisConfig()` and reads the saved config, so `CCS_IMAGE_ANALYSIS_BACKEND_ID` was correctly set to the mapped backend while the status object computed from the constant reported `native-read`. The env and the status disagreed about the same launch.

## Fix

Default to `getImageAnalysisConfig()` — the same source `getImageAnalysisHookEnv` already uses — so both agree. Callers passing an explicit config are unaffected.

## Verification

The new test fails on `dev` (`Expected: "claude"`, `Received: "gemini"`) and passes with the fix.

- `bun test tests/unit/utils/hooks/` — 30 pass
- image-analysis suites across hooks, web-server, management, commands — 72 pass
- `bun run typecheck` — clean
- `bun run format:check` — clean
- `bun run lint` — 0 errors (52 pre-existing `max-lines` warnings, unrelated files)
- `bun run test:fast` — 414 test files pass

`image-analysis-runtime-status-circular-dependency.test.ts` builds its own `requireMap`, so it needed the new module specifier added; its assertions are unchanged.

Manual check that the backend itself was never the problem — the same request the runtime builds, replayed against the remote CLIProxy provider route, reads the image correctly:

```
POST /api/provider/claude/v1/messages
-> "The image contains red and white colors."
```

## Notes

Both launch paths are fixed by the single default because they share this resolver. I did not change the call sites themselves, so any future caller that omits the config also gets the user's saved settings rather than the constant.

### DIFF
--- a/src/utils/hooks/image-analysis-runtime-status.ts
+++ b/src/utils/hooks/image-analysis-runtime-status.ts
@@ -16,10 +16,8 @@ import { getProxyTarget, type ProxyTarget } from '../../cliproxy/proxy/proxy-tar
 import { getProviderDisplayName, isCLIProxyProvider } from '../../cliproxy/provider-capabilities';
 import { isCliproxyRunning } from '../../cliproxy/services/stats-fetcher';
 import type { CLIProxyProvider } from '../../cliproxy/types';
-import {
-  DEFAULT_IMAGE_ANALYSIS_CONFIG,
-  type ImageAnalysisConfig,
-} from '../../config/unified-config-types';
+import { getImageAnalysisConfig } from '../../config/config-loader-facade';
+import { type ImageAnalysisConfig } from '../../config/unified-config-types';
 import {
   resolveImageAnalysisStatus,
   type ImageAnalysisResolutionContext,
@@ -191,9 +189,14 @@ export async function hydrateImageAnalysisRuntimeStatus(
 
 export async function resolveImageAnalysisRuntimeStatus(
   context: ImageAnalysisResolutionContext,
-  config: ImageAnalysisConfig = DEFAULT_IMAGE_ANALYSIS_CONFIG,
+  config?: ImageAnalysisConfig,
   deps: Partial<ImageAnalysisRuntimeStatusDeps> = {}
 ): Promise<ImageAnalysisStatus> {
-  const baseStatus = resolveImageAnalysisStatus(context, config);
+  // Fall back to the user's saved image_analysis config, not the built-in
+  // constant. Launch paths call this without an explicit config, and the
+  // constant carries empty profile_backends plus a gemini fallback_backend,
+  // so user mappings were dropped and every profile resolved to gemini.
+  const resolvedConfig = config ?? getImageAnalysisConfig();
+  const baseStatus = resolveImageAnalysisStatus(context, resolvedConfig);
   return hydrateImageAnalysisRuntimeStatus(baseStatus, deps);
 }
--- a/tests/unit/utils/hooks/image-analysis-runtime-status-circular-dependency.test.ts
+++ b/tests/unit/utils/hooks/image-analysis-runtime-status-circular-dependency.test.ts
@@ -89,8 +89,15 @@ describe('image-analysis-runtime-status circular dependency regression', () => {
       '../../cliproxy/services/stats-fetcher': {
         isCliproxyRunning: async () => true,
       },
-      '../../config/unified-config-types': {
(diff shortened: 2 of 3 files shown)


aaron-tsar:
CI shows `action_required` — the workflows are waiting on maintainer approval for a first-time fork contributor, so no jobs have run yet. Full local run in the meantime:

- `bun run test:fast` — 414 test files pass
- `bun run test:all` (slow bucket) — 81 test files pass
- `bun run typecheck` — clean
- `bun run format:check` — clean
- `bun run lint` — 0 errors (52 pre-existing `max-lines` warnings in unrelated files)

One note on `validate:ci-parity`: `test:runtime-matrix` fails locally for me, but it fails identically on a clean `dev` checkout with no changes applied — `Unable to resolve Node 22: npm error process terminated / signal SIGTERM`. That is my sandbox failing to fetch the pinned Node version, not a regression from this PR. Everything else in the parity gate passes.

Happy to rebase or adjust if you'd like the change shaped differently.

aaron-tsar:
Follow-up: verified the fix end-to-end on a real remote-CLIProxy setup, and found the impact is worse than the original description.

**The forced fallback can point at a provider that does not exist on the target.**

`DEFAULT_IMAGE_ANALYSIS_CONFIG` hardcodes `fallback_backend: 'gemini'`. On my remote CLIProxy, gemini is not configured at all:

```
POST /api/provider/gemini/v1/messages  model=gemini-3-flash-preview
-> {"error": "unknown provider for model gemini-3-flash-preview"}
```

`/v0/management/auth-files` lists only `claude` (3 files) and `codex` (7 files). So the user cannot resolve this by authenticating — there is no gemini provider on that server to authenticate against. `ccs gemini --auth` on the client does not help either: with `cliproxy_server.remote.enabled: true`, `resolveAuthReadiness` takes the `target.isRemote` branch and queries the remote server, never reading local token files.

That leaves no configuration-level workaround. Setting the env vars in the profile's `*.settings.json` does not work either, because `imageAnalysisEnv` is merged after `runtimeEnvVars` in `headless-executor.ts:470-482` and overwrites every `CCS_IMAGE_ANALYSIS_*` key unconditionally.

**Verification after the fix**

Installed the patched build globally and launched the affected profile (a deepseek settings profile mapped to `claude` via `profile_backends`, model has no vision support):

Before — every launch:
```
[i] Google Gemini auth is missing. Run "ccs gemini --auth" to enable image analysis.
    This delegation will use native Read.
```

After — warning gone, and the profile reads images correctly through the mapped backend:
```
$ ccs deepseek -p "Use ImageAnalysis on ccs-verify.png and tell me only what color it is."
Blue.
```
(test image was solid `#0080ff`)

Also worth flagging for a possible separate issue: `/api/provider/<anything>/v1/models` returns HTTP 200 with an identical Claude model list for any provider segment, including nonsense ones like `xyz-fake`. It is not a usable liveness signal for a provider — I initially misread it as one.

PR #1718, opened 2026-08-23
fix(shared-manager): publish adopted settings by replacement

sgaluza:
## Problem

Adopting a diverged `settings.json` can destroy the user's live settings.

`adoptDivergedFileContent` moves the canonical file aside with `rename()` (`diverged-file-adopter.ts:296` on `dev`) and leaves `~/.claude/settings.json` missing until publication, roughly 100 ms later — two full file copies (`publishBackupNoReplace`, `publishAdoptedRecoveryNoReplace`) run inside that window. Anything starting in it finds no file and seeds an empty placeholder: Claude Code does this on startup, and so does a second `ccs` through `shared-dir-linker.ts:127`. Publication then fails with `EEXIST`, because `atomicWriteFile` publishes through `link()`, which is no-replace. The rollback in `restoreCanonicalClaim` hits the same `EEXIST`, publishes a backup and unlinks the claim — destroying the inode that held the only remaining copy. The live path is left holding the placeholder, and the user gets no indication that anything happened.

We hit this on 2026-08-20: a 5580-byte `settings.json` with 168 `permissions.allow` rules, `model`, `hooks`, `theme` and `enabledPlugins` became `{}` — 3 bytes. Every instance shares that file through the managed symlink chain, so it went everywhere at once. The content survived only in `.bak-ccs-adopt` and `.ccs-adopted-recovery` sidecars, and recovering it meant matching them up by hand. The same code path had run correctly days earlier, which is what makes it a race rather than a plain bug.

Two concurrent `ccs` runs are enough to reproduce it; Claude Code is not required.

## Fix

Publish by replacement. A fully written, fsynced temp file is renamed over the canonical inode, so the path always holds a regular file and no placeholder can be seeded. The claim on the canonical file is gone entirely, along with `restoreCanonicalClaim`.

A compare-and-swap guard on `(ino, mtime, size)` runs immediately before the rename and refuses to publish when the canonical inode changed since it was read, so a writer that got there first is still never clobbered — that guarantee now costs nothing, because only a genuinely concurrent write can land there. The guard is read-then-act, not atomic; POSIX offers no compare-and-swap rename, so the window shrinks from ~100 ms to two adjacent syscalls rather than closing. That limit is stated in the code so it is not mistaken for strict atomicity, and the pre-image backup — published before the replacement — keeps even that outcome recoverable.

`recoverOrphanedCanonicalClaim` is kept: claims written by older versions may still be on disk.

Follow-on cleanups in the same area: the two identical sidecar publishers fold into one helper, and the duplicated durable-temp-write block behind both publishers is extracted (`writeDurableTempFile`, `discardTempFile`).

## Reversed test

`preserves a canonical write that lands during no-replace publication` pinned the incident's outcome as intended behaviour — the foreign write survives, the previous canonical content lives only in `.bak-ccs-adopt`. It is rewritten as `refuses to publish when the canonical file changes before publication`. The intent it encoded — never clobber a writer that got there first — is preserved and now enforced by the CAS guard. Three more tests keyed on the `.ccs-canonical-claim-` path were moved to injection points that still exist.

## Test plan

- Two new cases in `tests/unit/shared-manager.test.ts` reproduce the race deterministically: a foreign writer seeds an empty placeholder the moment the canonical path is observed missing, once as Claude Code (`{}` with a trailing newline) and once as a second `ccs` (the 2-byte seed from `shared-dir-linker`). Both fail on `dev` and pass here. They also assert the writer never fired at all, which is the invariant the fix establishes.
- `tests/unit/shared-manager.test.ts` — 41 pass / 0 fail.
- `bun test tests/unit` — 3842 pass, 24 skip.
- `bun run typecheck`, `bun run lint`, `bun run format:check` — clean.
- `scripts/ci-parity-gate.sh` — passes; `docs/reports/hardening-inventory.{json,md}` regenerated as the gate requires. The adopter drops from 36 to 29 sync-fs call sites after the dedup.

### DIFF
--- a/docs/reports/hardening-inventory.json
+++ b/docs/reports/hardening-inventory.json
@@ -1,15 +1,16 @@
 {
   "scope": "src/**/*.{ts,tsx,js,jsx,mjs,cjs}",
   "syncFs": {
-    "totalOccurrences": 2519,
+    "totalOccurrences": 2516,
     "filesAffected": 263,
-    "hotpathOccurrences": 1193,
+    "hotpathOccurrences": 1190,
     "hotpathFilesAffected": 156,
     "topHotpathFiles": [
       {
         "file": "src/management/shared-manager/diverged-file-adopter.ts",
-        "count": 39,
+        "count": 36,
         "calls": [
+          "chmodSync",
           "closeSync",
           "fsyncSync",
           "linkSync",
@@ -245,9 +246,24 @@
         "markers": []
       },
       {
-        "file": "src/management/shared-manager/diverged-file-adopter.ts",
-        "count": 39,
+        "file": "src/cliproxy/executor/__tests__/variant-port-integration.test.js",
+        "count": 36,
         "calls": [
+          "existsSync",
+          "mkdirSync",
+          "readdirSync",
+          "readFileSync",
+          "rmSync",
+          "unlinkSync",
+          "writeFileSync"
+        ],
+        "markers": []
+      },
+      {
+        "file": "src/management/shared-manager/diverged-file-adopter.ts",
+        "count": 36,
+        "calls": [
+          "chmodSync",
           "closeSync",
           "fsyncSync",
           "linkSync",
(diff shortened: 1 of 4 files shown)


kaitranntt (approved):
### Summary
Fixes settings destruction race condition in `adoptDivergedFileContent` by publishing adopted content via atomic replacement (`renameSync`) instead of moving the canonical file aside first. Introduces a compare-and-swap (CAS) guard on inode/mtime/size immediately prior to replacement and durable pre-image sidecar backups.

### Risk Level
Low

### Findings
None. All 41 shared-manager tests (including deterministic concurrent placeholder writer race reproductions) and full repository CI parity checks pass cleanly with 0 errors.

### Verdict
**Approve**

*Posted by the installed review-pr skill at 2026-08-25T18:50:10Z*

PR #1721, opened 2026-08-25
fix(websearch): filter disabled legacy CLI probes and skip unused version fetching (#1716)

### LINKED ISSUE #1716: bug: launch path probes every legacy WebSearch CLI, including disabled ones, and discards the version it fetched
**Affected area:** CLI runtime (WebSearch provider probing on the launch path)

## What broke?

On every launch of a settings/API profile, `displayWebSearchStatus()` checks all four legacy CLI providers (agy, gemini, grok, opencode) by spawning subprocesses. It does this before checking whether those providers are enabled in config. Two problems fall out of this:

1. **Disabled providers are probed anyway.** `getLegacyProviderStatuses()` calls all four status functions unconditionally. The `enabled` filter only runs later, in `buildWebSearchReadiness()`. By then the subprocesses have already run and been waited on.
2. **A version string is fetched and then thrown away.** For `opencode`, `gemini`, and `grok`, a second subprocess asks the CLI for its version. Nothing on the launch path ever shows that version. Only `agy` displays it.

On my machine this costs about 1.6 seconds per launch, most of it for a provider I have set to `enabled: false`. Anyone who has one of these CLIs on `PATH` but disabled in config pays the same cost.

### Where the time goes

`ccs glm --version`, idle machine. Every row except `where gemini` is a provider set to `enabled: false`:

```
where agy           ██                                    159 ms    disabled
where gemini        ██                                    149 ms    ENABLED
where grok          ██                                    153 ms    disabled
where opencode      ██                                    164 ms    disabled
opencode --version  ████████████████████████████████     1152 ms    disabled  ← 65% of the block
                    ─────────────────────────────────────────────
                    total                                1777 ms
                    serving an enabled provider           149 ms    (8%)
```

92% of that block is spent on providers whose results are filtered out a moment later, inside the same call.

## Actual behavior (with code references)

Line numbers are against `main` as of v8.9.0.

**1. Probes run before the enabled filter.** `src/utils/websearch/status.ts:116-120`:

```ts
function getLegacyProviderStatuses(wsConfig: WebSearchConfigSnapshot): WebSearchCliInfo[] {
  const agyStatus = getAgyCliStatus();           // spawns `where agy`
  const geminiStatus = getGeminiCliStatus();     // spawns `where gemini`
  const grokStatus = getGrokCliStatus();         // spawns `where grok`
  const opencodeStatus = getOpenCodeCliStatus(); // spawns `where opencode` + `opencode --version`
```

`wsConfig` is already a parameter here (added by commit abc3002 for #1648), but it is only used later to fill the `enabled:` field of each returned object (for example `status.ts:166`).

The filter runs downstream, in `buildWebSearchReadiness()` at `status.ts:328`:

```ts
const enabledProviders = providers.filter((provider) => provider.enabled);
```

That is: after every subprocess has already run.

**2. The version is fetched and thrown away.** Compare the `agy` entry, which shows its version (`status.ts:136-140`):

```ts
      detail: agyStatus.installed
        ? agyStatus.version
          ? `Installed (${agyStatus.version})`   // agy shows its version
          : 'Installed'
        : 'Not installed',
```

with the `opencode` entry, which does not (`status.ts:162-174`):

```ts
      id: 'opencode',
      kind: 'legacy-cli',
      name: 'OpenCode',
      command: 'opencode',
      enabled: wsConfig.providers?.opencode?.enabled ?? false,
      available: opencodeStatus.installed,     // comes from `where` alone
      version: opencodeStatus.version ?? null, // comes from `opencode --version`
      installCommand: 'curl -fsSL https://opencode.ai/install | bash',
      docsUrl: 'https://github.com/sst/opencode',
      requiresApiKey: false,
      description: 'Optional legacy LLM fallback via OpenCode.',
      detail: opencodeStatus.installed ? 'Installed' : 'Not installed',  // version unused
```

`gemini`'s `detail` shows auth state and `grok`'s shows whether an API key is set. Neither shows the version.

The status line printed at launch is built from `name` and `detail` only (`status.ts:334` and `:342`):

```ts
      message: `Ready (${readyProviders.map((provider) => provider.name).join(' + ')})`,
// or
      message: enabledProviders
        .map((provider) => `${provider.name}: ${provider.detail}`)
        .join(' | '),
```

Searching the codebase for readers of `provider.version` finds the four places that set it, plus agy's `detail`. On the launch path, the version of opencode, gemini, and grok is written but never read:

```mermaid
flowchart LR
  A["agy --version"] --> B["agyStatus.version"]
  B --> C["detail: 'Installed (x.y.z)'<br/>status.ts:138"]
  C --> D["readiness message"]

  E["opencode --version<br/>1152 ms"] --> F["opencodeStatus.version"]
  F --> G["provider.version<br/>status.ts:168"]
  G --> H["never read"]

  style H fill:#5a1e1e,stroke:#c0392b,color:#fff
  style E fill:#5a1e1e,stroke:#c0392b,color:#fff
```

### Why the wasted call is so expensive

`opencode --version` is not a cheap call. Current opencode ships as a single-file executable built with Bun's `--compile`: the whole Bun runtime plus the application, bundled into one binary.

```
C:\ProgramData\chocolatey\lib\opencode\tools\opencode.exe
size          : 152.65 MB
product name  : Bun      (from the executable's Windows file properties)
Bun version   : 1.3.9
```

There is no fast path for `--version`: the runtime boots and evaluates the whole bundle before it reads its command-line arguments. Verified by passing a flag that does not exist:

```
opencode --zzz-nonexistent-flag   →  1409 ms, and it still renders the full ASCII logo
```

Same cost as `--version`, so the time is startup, not version lookup.

Measured on this machine:

| | Time |
|---|---:|
| `where.exe` (baseline process spawn) | ~115 ms |
| `claude --version` | ~222 ms |
| `opencode --version` (idle machine) | ~1010 ms |
| `opencode --version` (machine at 88% CPU) | ~5450 ms |

Five warm runs in a row on an idle machine: 1082 / 1150 / 1009 / 1013 / 983 ms. Flat, so this is not first-run disk caching. The startup work repeats in full on every call.

Note the load sensitivity: the cost grows about 5x when the CPU is busy, so the waste is largest exactly when the machine can least afford it.

## Expected behavior

Providers that are disabled in config are not probed, and the launch path does not pay for a version string it never shows.

The fix is a change of ordering. The `enabled` filter already exists; it just runs after the subprocesses instead of before them:

```mermaid
flowchart TB
  subgraph NOW["BEFORE - probe, then filter"]
    direction TB
    N1["getLegacyProviderStatuses(wsConfig)"]
    N1 -->|"spawn 159ms"| N2["where agy"]
    N1 -->|"spawn 149ms"| N3["where gemini"]
    N1 -->|"spawn 153ms"| N4["where grok"]
    N1 -->|"spawn 1316ms"| N5["where opencode<br/>+ opencode --version"]
    N2 --> N6["build 4 provider objects"]
    N3 --> N6
    N4 --> N6
    N5 --> N6
    N6 -->|"discards 3 of 4"| N7["filter by enabled<br/>status.ts:328"]
    N7 --> N8["'Gemini CLI: Not installed'"]
  end
```

```mermaid
flowchart TB
  subgraph FIX["AFTER - filter, then probe"]
    direction TB
    F1["getLegacyProviderStatuses(wsConfig)"]
    F1 -->|"wsConfig already in scope"| F2["filter by enabled"]
    F2 -->|"spawn 149ms"| F3["where gemini"]
    F2 -.->|"not spawned"| F4["agy, grok, opencode"]
    F3 --> F5["build 1 provider object"]
    F5 --> F6["'Gemini CLI: Not installed'<br/>identical output"]
  end
```

Same printed output, 1628 ms less work.

Concretely:

1. In `getLegacyProviderStatuses()`, skip the probe for providers that are disabled. `wsConfig` is already in scope.
2. Make version fetching opt-in, so the dashboard keeps it and the launch path skips it where the version is never shown:

```ts
getWebSearchCliProviders(wsConfig, { includeVersions: false })  // launch path
getWebSearchCliProviders(wsConfig, { includeVersions: true })   // web-server
```

One exception to state plainly: agy's `detail` does show its version on the launch path ("Installed (x.y.z)"), so the version skip applies to opencode, gemini, and grok only. agy either keeps its version call, or its launch-path detail drops the version (a small visible change, maintainer's call). Change 1 has no such exception: a disabled provider's entire result is discarded, so skipping its probe cannot change output.

The dashboard (`web-server/routes/websearch-routes.js`, `web-server/health/websearch-checks.js`) is the only other consumer of `getWebSearchCliProviders()`. That is a long-running server where the probe cost is harmless and showing versions is genuinely useful, so the two use cases separate cleanly.

## Reproduction steps

1. In `~/.ccs/config.yaml`, set `websearch.enabled: true` and `websearch.providers.opencode.enabled: false`.
2. Install opencode so it resolves on `PATH` (the standalone build, for example `choco install opencode`).
3. Launch any settings/API profile: `ccs <profile> --version`.
4. Watch the spawned processes: `opencode --version` runs (about 1 second), while the printed status line never mentions opencode. It was filtered out downstream.

Trace on my machine, `ccs glm --version`:

```
where agy               159 ms
where gemini            149 ms
where grok              153 ms
where opencode          164 ms
opencode --version     1152 ms   ← disabled provider, result never displayed
                      ────────
                       1777 ms
```

Printed result: `WebSearch: Gemini CLI: Not installed (retired - use Antigravity)`

## CCS version

8.9.0 (npm `@kaitranntt/ccs`)

## Operating system

Windows

## Environment details

- Windows 11 (10.0.26200)
- Node.js v24.12.0
- Claude Code 2.1.227
- opencode 1.1.65, installed via Chocolatey (Bun 1.3.9 single-file build, 152.65 MB)

## Before submitting

- [x] I searched existing issues first.
- [x] I removed secrets and private data from logs/screenshots.
- [x] I can still reproduce this on the latest released or dev build.

---

## Related, but separable: repeated `claude` binary detection

The account-profile launch path asks Windows where `claude` is installed three times in one process, then runs the `claude` binary twice more to ask it questions, before finally launching it for real:

```
ccs <account-profile> --version

where.exe claude       ~236 ms   from profile-resolver
where.exe claude       ~236 ms   inside getClaudeCliVersion()
where.exe claude       ~236 ms   inside getClaudeAuthStatus()
claude --version      ~1200 ms   subprocess
claude auth status    ~2400 ms   subprocess
                      ────────
                       ~4300 ms  before the real launch starts
```

`detectClaudeCli()` (`src/utils/claude-detector.ts`) never stores its result, and every answer above is identical on every run until the binary moves or updates.

I prototyped a fix: a small on-disk cache (about 400 bytes) for the binary path and version, keyed by proof the answer is still valid (a hash of `PATH` for the location; the binary's modified-time plus size for the version). Login state can change at any time, so `getClaudeAuthStatus()` stays uncached.

Result, measured on the same tree with the patch toggled by an env var, runs alternated between the two settings:

| Probe | uncached | cached | saved |
|---|---:|---:|---:|
| `detectClaudeCli` | 545 ms | 3 ms | 542 |
| `getClaudeCliVersion` | 1208 ms | 4 ms | 1204 |
| `getOpenCodeCliStatus` | 5457 ms | 1 ms | 5456 |
| `getClaudeAuthStatus` *(uncached on purpose)* | 5233 ms | 2354 ms | — |

End to end:

| | uncached | cached | saved |
|---|---:|---:|---:|
| `ccs glm --version` | 4766 ms | 3149 ms | 1617 ms (34%) |
| `ccs claude-pro --version` (account profile: official Claude subscription) | 5571 ms | 3805 ms | 1766 ms (32%) |

Output verified byte-identical with the cache on and off, exit 0 in both. (The machine was at 88% CPU during these runs, so absolute numbers are inflated; the comparison held across all iterations.)

I am happy to open a PR for either or both of these if the approach sounds right.

kaitranntt:
## Summary
Optimizes the launch-path WebSearch readiness check by filtering out disabled legacy CLI providers before probing, and skipping unused version subprocess calls (`opencode --version`, `gemini --version`, `grok --version`) when rendering launch readiness.

## Changes
- `src/utils/websearch/status.ts`:
  - `getLegacyProviderStatuses`: Checks `wsConfig.providers?.[id]?.enabled` before executing subprocess probes. Returns uninstalled placeholder if disabled.
  - `getWebSearchCliProviders`: Supports optional `includeVersions` parameter.
  - `getWebSearchReadiness`: Passes `{ includeVersions: false }` on the launch path.
- `src/utils/websearch/agy.ts`, `gemini-cli.ts`, `grok-cli.ts`, `opencode-cli.ts`:
  - Added optional `options?: { fetchVersion?: boolean }` to status detection functions, skipping expensive `--version` subprocess calls when `fetchVersion: false`.
- `tests/unit/utils/websearch/status.test.ts`:
  - Added unit tests asserting disabled legacy CLI providers do not trigger subprocess probes.
  - Added unit tests asserting `includeVersions: false` passes `fetchVersion: false`.

Closes #1716

### DIFF
--- a/src/utils/websearch/agy.ts
+++ b/src/utils/websearch/agy.ts
@@ -22,35 +22,47 @@ let agyCliCache: AgyCliStatus | null = null;
  *
  * @returns Antigravity CLI status with path and version
  */
-export function getAgyCliStatus(): AgyCliStatus {
-  // Return cached result if available
-  if (agyCliCache) {
+export function getAgyCliStatus(options?: { fetchVersion?: boolean }): AgyCliStatus {
+  const fetchVersion = options?.fetchVersion !== false;
+
+  if (
+    agyCliCache &&
+    (!fetchVersion || agyCliCache.version !== undefined || !agyCliCache.installed)
+  ) {
+    if (!fetchVersion) {
+      return { ...agyCliCache, version: undefined };
+    }
     return agyCliCache;
   }
 
-  const result: AgyCliStatus = {
-    installed: false,
-    path: undefined,
-    version: undefined,
-  };
+  const result: AgyCliStatus = agyCliCache
+    ? { ...agyCliCache }
+    : {
+        installed: false,
+        path: undefined,
+        version: undefined,
+      };
 
   try {
-    const isWindows = process.platform === 'win32';
-    const whichCmd = isWindows ? 'where agy' : 'which agy';
+    if (!agyCliCache) {
+      const isWindows = process.platform === 'win32';
+      const whichCmd = isWindows ? 'where agy' : 'which agy';
 
-    const pathResult = execSync(whichCmd, {
-      encoding: 'utf8',
-      timeout: 5000,
-      stdio: ['pipe', 'pipe', 'pipe'],
-    });
+      const pathResult = execSync(whichCmd, {
+        encoding: 'utf8',
+        timeout: 5000,
+        stdio: ['pipe', 'pipe', 'pipe'],
+      });
 
-    const agyPath = pathResult.trim().split('\n')[0]; // First result on Windows
+      const agyPath = pathResult.trim().split('\n')[0];
 
-    if (agyPath) {
-      result.installed = true;
-      result.path = agyPath;
+      if (agyPath) {
+        result.installed = true;
+        result.path = agyPath;
+      }
+    }
 
-      // Try to get version
+    if (result.installed && fetchVersion && result.version === undefined) {
       try {
         const versionResult = execSync('agy --version', {
           encoding: 'utf8',
@@ -59,17 +71,17 @@ export function getAgyCliStatus(): AgyCliStatus {
         });
         result.version = versionResult.trim();
       } catch {
-        // Version check failed, but CLI is installed
         result.version = 'unknown';
       }
+    } else if (!fetchVersion) {
+      result.version = undefined;
     }
   } catch {
     // Command not found - Antigravity CLI not installed
   }
 
-  // Cache result
   agyCliCache = result;
-  return result;
+  return fetchVersion ? result : { ...result, version: undefined };
 }
 
 /**
--- a/src/utils/websearch/gemini-cli.ts
+++ b/src/utils/websearch/gemini-cli.ts
@@ -23,35 +23,47 @@ let geminiCliCache: GeminiCliStatus | null = null;
  *
  * @returns Gemini CLI status with path and version
  */
-export function getGeminiCliStatus(): GeminiCliStatus {
-  // Return cached result if available
-  if (geminiCliCache) {
+export function getGeminiCliStatus(options?: { fetchVersion?: boolean }): GeminiCliStatus {
+  const fetchVersion = options?.fetchVersion !== false;
+
+  if (
+    geminiCliCache &&
+    (!fetchVersion || geminiCliCache.version !== undefined || !geminiCliCache.installed)
+  ) {
+    if (!fetchVersion) {
+      return { ...geminiCliCache, version: undefined };
+    }
     return geminiCliCache;
   }
 
-  const result: GeminiCliStatus = {
-    installed: false,
-    path: undefined,
-    version: undefined,
-  };
+  const result: GeminiCliStatus = geminiCliCache
+    ? { ...geminiCliCache }
+    : {
+        installed: false,
+        path: undefined,
+        version: undefined,
+      };
 
   try {
-    const isWindows = process.platform === 'win32';
-    const whichCmd = isWindows ? 'where gemini' : 'which gemini';
+    if (!geminiCliCache) {
+      const isWindows = process.platform === 'win32';
+      const whichCmd = isWindows ? 'where gemini' : 'which gemini';
 
-    const pathResult = execSync(whichCmd, {
-      encoding: 'utf8',
-      timeout: 5000,
-      stdio: ['pipe', 'pipe', 'pipe'],
-    });
+      const pathResult = execSync(whichCmd, {
+        encoding: 'utf8',
+        timeout: 5000,
+        stdio: ['pipe', 'pipe', 'pipe'],
+      });
 
-    const geminiPath = pathResult.trim().split('\n')[0]; // First result on Windows
+      const geminiPath = pathResult.trim().split('\n')[0];
 
-    if (geminiPath) {
-      result.installed = true;
-      result.path = geminiPath;
+      if (geminiPath) {
+        result.installed = true;
+        result.path = geminiPath;
+      }
+    }
 
-      // Try to get version
+    if (result.installed && fetchVersion && result.version === undefined) {
       try {
         const versionResult = execSync('gemini --version', {
           encoding: 'utf8',
@@ -60,17 +72,17 @@ export function getGeminiCliStatus(): GeminiCliStatus {
         });
         result.version = versionResult.trim();
       } catch {
-        // Version check failed, but CLI is installed
         result.version = 'unknown';
       }
+    } else if (!fetchVersion) {
+      result.version = undefined;
     }
   } catch {
     // Command not found - Gemini CLI not installed
   }
 
-  // Cache result
   geminiCliCache = result;
-  return result;
+  return fetchVersion ? result : { ...result, version: undefined };
 }
 
 /**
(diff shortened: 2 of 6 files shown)


kaitranntt (commented):
### Summary
Optimizes launch-path WebSearch readiness check by skipping disabled legacy CLI probes before subprocess spawning, keeping `agy` version resolution, and making version lookup opt-in (`includeVersions: false`) on the launch path.

### Risk Level
Low

### Findings
- Startup latency reduced by eliminating unused `--version` spawns (e.g. `opencode --version`).
- Truthful `Disabled` detail rendered for unprobed disabled providers.
- Dashboard routes continue to probe full status with versions when enabled.
- CI is 100% green.

### Verdict
**Approve**

*Posted by the installed review-pr skill at 2026-08-25T20:02:32Z*

